PR #9, opened 2019-05-13
Add E2E tests

nikhil-thomas:
Add e2e tests

Add a test document **docs/tests.md**

### DIFF
--- a/deploy/role_binding.yaml
+++ b/deploy/role_binding.yaml
@@ -6,7 +6,7 @@ subjects:
 - kind: ServiceAccount
   name: openshift-pipelines-operator
   # Replace this with the namespace the operator is deployed in.
-  namespace: openshift-pipelines-operator
+  namespace: dev-openshift-pipelines-operator
 roleRef:
   kind: ClusterRole
   name: openshift-pipelines-operator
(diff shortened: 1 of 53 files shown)


deploy/role_binding.yaml
```diff
@@ -6,7 +6,7 @@ subjects:
 - kind: ServiceAccount
   name: openshift-pipelines-operator
   # Replace this with the namespace the operator is deployed in.
-  namespace: tekton-pipelines
+  namespace: dev-openshift-pipelines-operator
```
sthaha: why has the namespace changed?

nikhil-thomas: Openshift will no let you name a namespace with a string starting with 'openshift'

nikhil-thomas: we need a namespace to run the tests. The operator should have necessary permissions in that namespace. so i set the namespace as `dev-openshift-pipelines-operator` in the role_binding and used it in the run test command
```
operator-sdk test local ./test/e2e --namespace dev-openshift-pipelines-operator --debug
```

nikhil-thomas: while using the operator with OLM the permissions are set from CSV not from role_binding.yaml

Openshift will no let you name a namespace with a string starting with 'openshift'. Hence the `dev-` prefix

PR #14, opened 2019-05-16
addon support

akihikokuroda:
The tektoncd has experimental repository that has additional / optional components for the tektoncd pipelines.  I would like to add mechanism to install these components optionally to the operator.   This PR has change to make the optional component included and installed with the following steps.

The optional component is included bu placing the necessary config yaml files for the component under deploy/resource/"component name"/ directory

The optional component is installed by adding component name to the spec.addons list in the /deploy/crds/openshift_v1alpha1_install_cr.yaml like this.
```
apiVersion: tekton.dev/v1alpha1
kind: Install
metadata:
  name: example-install
  namespace: openshift-pipelines-operator
spec:
  addons:
  - componentName1
  - componentName2
```


akihikokuroda:
We can mount the config map that has the yaml files for the component in the /deploy/resource/"compoment name" directory so that we don't need to rebuild the operator image to integrate the optional components.

akihikokuroda:
OK. I'll look at how it works and add some.

akihikokuroda:
I added the test.  Unfortunately I couldn't add the test just by changing the go test.  I had to patch the deployment of the operator but the operator deployment definition is read before the go test is called.  So I added the script file that do the pretest / post test work in addition to invoke the test.  
Here is the way to execute the test.
```
test/resources/e2etest.sh <image name>
```

pkg/controller/install/install_controller.go
```diff
@@ -166,9 +169,15 @@ func (r *ReconcileInstall) install(instance *tektonv1alpha1.Install) error {
...
-
 	r.manifest.Filter(filters...)
-	return r.manifest.ApplyAll()
+	rc := r.manifest.ApplyAll()
+
+	for _, path := range instance.Spec.AddOns {
+		extension := mf.NewYamlManifest(filepath.Join("deploy", "resources", path), true, r.config)
+		extension.Filter(filters...)
+		extension.ApplyAll()
+	}
```
nikhil-thomas: Will the addon resources be deleted when we remove the instance of `install` CRD. Resources in `r.manifest` are deleted (line 138) when the CRD instance is removed.

test/resources/addon1.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: addon1
+  namespace: openshift-pipelines-operator
```
nikhil-thomas: the namespace here should be `tekton-pipelines`
as the `tekton-pipelines-controller`, `tekton-pipelines-webhook` and `addon1` expected to appear in `tekton-pipelines` namespace

test/resources/addon1.yaml
```diff
@@ -0,0 +1,25 @@
...
+    spec:
+      replicas: 1
+      template:
+        metadata:
+          labels:
+            app: nginx
+        spec:
+          containers:
+          - name: nginx
+            image: nginx:1.15.12
```
nikhil-thomas: nit.
shall we use a lighter image, like `1.15-alpine`

pkg/controller/install/install_controller.go
```diff
@@ -166,9 +169,15 @@ func (r *ReconcileInstall) install(instance *tektonv1alpha1.Install) error {
...
 	}
-
 	r.manifest.Filter(filters...)
-	return r.manifest.ApplyAll()
+	rc := r.manifest.ApplyAll()
+
+	for _, path := range instance.Spec.AddOns {
+		extension := mf.NewYamlManifest(filepath.Join("deploy", "resources", path), true, r.config)
+		extension.Filter(filters...)
+		extension.ApplyAll()
```
nikhil-thomas: receive the return value here and return it if `err != nil`

test/resources/addon1.yaml
akihikokuroda: I'll change to alpine

pkg/controller/install/install_controller.go
akihikokuroda: I'll add it.

pkg/controller/install/install_controller.go
akihikokuroda: Yes, they will.  The addon resources are owned by the install instance.  When the install instance is deleted, the addon resources owned by the instance are deleted automatically.

nikhil-thomas:
@akihikokuroda first of all, i apologize for the delay.
We had to take care of some critical patches as one of our core dependencies had to be updated.
it had some breaking changes. (https://github.com/jcrossley3/manifestival)

We have always kept it constraint in Gopkg.toml using the commit hash. so our code will not break even if there are breaking changes in that packages. This time we had to update the package as we could make use of some fixes/features in `manifestival`.

I rebased your branch locally and ran the tests. It works.
you can find the changes here https://github.com/nikhil-thomas/tektoncd-pipeline-operator/tree/akihikokuroda-addons-rebase

Could you take a look at the changes and update your Pull Request.
Thank you 100 times for your contribution :)

akihikokuroda:
I’ll rebase as soon as I come back from my vacation.  Thanks!

nikhil-thomas:
> I’ll rebase as soon as I come back from my vacation. Thanks!
:+1:

akihikokuroda:
I finished the rebase.

akihikokuroda:
I still have to update the test.

sthaha:
@akihikokuroda I have closed this PR in favour of #23 , please reopen and rebase if this pr needs to be merged instead.

PR #29, opened 2019-06-12
Fix CI build

chmouel:
Fix CI to pas green by disabling some features that needs to be implemented first,

/cc @nikhil-thomas @sthaha @vdemeester


chmouel:
/hold cancel

ok we just need to have the move to manifests files done by @nikhil-thomas and this should get ready to pass,

PR #35, opened 2019-06-13
Changed the container runtime references

knobunc:
Changed from docker -> container runtime/image to be more generic as we move to support more runtimes.


nikhil-thomas:
Thanks

PR #47, opened 2019-06-17
Make tekton-pipelines as namespace

vincent-pli:
We expect `tekton-pipelines` as the name spaces of `tekton pipelines`.
So the commented line is no needed.

### DIFF
--- a/pkg/controller/install/install_controller.go
+++ b/pkg/controller/install/install_controller.go
@@ -179,7 +179,7 @@ func (r *ReconcileInstall) Reconcile(request reconcile.Request) (reconcile.Resul
 func (r *ReconcileInstall) install(instance *tektonv1alpha1.Install) error {
 	tfs := []mf.Transformer{
 		mf.InjectOwner(instance),
-		mf.InjectNamespace(instance.GetNamespace()),
+//		mf.InjectNamespace(instance.GetNamespace()),
 	}
 
 	err := r.manifest.Transform(tfs...)


nikhil-thomas:
This is a feature.
You coluld make `tekton-pipelines` available in the name spaces of tekton pipelines by creating the namespace 'tekton-pipelines' first and install the operator in that namespace.

sthaha:
> We expect `tekton-pipelines` as the name spaces of `tekton pipelines`.
> So the commented line is no needed.

is that a requirement? is that namespace hardcoded into the pipeline controllers?

vincent-pli:
There is `namespace` defined in the `deploy/resources` like this:
```
apiVersion: v1
kind: Namespace
metadata:
  name: tekton-pipelines
```
So I think we expect install `tekton-pipelines` in namespace `tekton-pipelines`
So need comment this code, or will be install in namespace `openshift-pipelines-operator`

sthaha:
> There is `namespace` defined in the `deploy/resources` like this:
> 
> ```
> apiVersion: v1
> kind: Namespace
> metadata:
>   name: tekton-pipelines
> ```
> 
> So I think we expect install `tekton-pipelines` in namespace `tekton-pipelines`
> So need comment this code, or will be install in namespace `openshift-pipelines-operator`


I think we have a difference in requirement on how the operator should behave. On OpenShift when you install the operator to `foobar` namespace, the assumption is that all `tekton-pipeline` components also get installed to the **same** namespace `foobar` (i.e. no magic). This behaviour involves less magic to `user` / `admin` who installed pipeline. 

In the case where `operator` creates `tekton-pipelines` (magically) they may be surprised to discover a `tekton-pipelines` namespace which they didn't create.

vincent-pli:
Fine, agree with you, no magic 😄

PR #48, opened 2019-06-17
Add permission to avoid exception

vincent-pli:
When delete `Install` the permission denied exception will raise in logs


nikhil-thomas:
Thank you for the patch.

vincent-pli:
Cool!

nikhil-thomas:
Thanks

nikhil-thomas:
could you update the permission in
deploy/role.yaml as well

vincent-pli:
The permission in `deploy/role.yaml` is right, no need to modify 😄

sthaha:
@vincent-pli now that #66 is merged, is this patch still needed?

sthaha:
@vincent-pli closing this since we have this running and we have the operator upstream.

PR #51, opened 2019-06-18
[TEST] Fix failing tests in CI

nikhil-thomas:
port 8383 was being used as metrics port in operator. Operator fails to run
when it cannot bind to port 8383.

This patch set metrics port reference in cmd/manager/main.go to 18383

E2E tests were being timedout because the operator was not
getting started properly.


cmd/manager/main.go
```diff
@@ -26,7 +26,7 @@ import (
 // Change below variables to serve metrics on different host or port.
 var (
 	metricsHost       = "0.0.0.0"
-	metricsPort int32 = 8383
+	metricsPort int32 = 18383
```
vdemeester: oh interesting :angel:

nikhil-thomas: > Let's see if it gets green

:pray:

nikhil-thomas:
> Very weird that it fails, did you check the version? because it works locally

i did. In the last run operator-sdk cli is version 0.8.1
and the operator-sdk pkg dependency in test code is also v0.8.1

right now im trying to run the operator process itslef without the test
`operator-sdk up local --namespace $(TEST_NAMESPACE) &  sleep 180; kill $!`

instead of

`operator-sdk test local ./test/e2e --namespace $(TEST_NAMESPACE) --up-local --debug --go-test-flags "-v -timeout=15m"`

i hope this might shed some light. if it works

nikhil-thomas:
@chmouel 
https://github.com/openshift/tektoncd-pipeline-operator/pull/51/commits/9fd737963b04a0ff393547c03a8fac32ffa49bb7

This passed as i used an image known to be working. 
So technically we can use image built by ci (ie, using `--image` flag instead of `--up-local` flag in test)

I tried it did not work 
https://github.com/openshift/tektoncd-pipeline-operator/pull/51/commits/371c72e3a83f7448da3bbcb024b103cafbe1e56b

nikhil-thomas:
but wait i think i missed something, the --verbose flag (which breaks build) i will try the above step again without --verbose

nikhil-thomas:
fixed in https://github.com/openshift/tektoncd-pipeline-operator/pull/71
This pr is obsolete. Hence closing

PR #63, opened 2019-06-25
Removes unwanted CI configuration

hrishin:
Current CI config has unwanted things which create
some confusions.
This patch removes wanted step `make build`  target
(OLM manifest generation) and drops extra docrerfiles.


hrishin:
closing this PR, I guess we need it in case of nightly.

PR #65, opened 2019-07-02
Changed install location to tekton-pipelines for Minishift

CarolynMabbott:
Changed install location to tekton-pipelines for minishift to get install to work with the Tekton dashboard as well as updating the instructions for installing on minishift 

References issue https://github.com/tektoncd/dashboard/pull/236

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,9 +108,12 @@ cd $GOPATH/src/github.com/operator-framework/operator-lifecycle-manager
 
 1. Change directory to `${GOPATH}/src/github.com/openshift/tektoncd-pipeline-operator`
 
-1. Create `openshift-pipelines-operator` namespace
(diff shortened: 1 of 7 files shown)


nikhil-thomas:
@CarolynMabbott why is it necessary to change the namespace to `tekton-pipelines`.

sthaha:
@CarolynMabbott I understand there may be a reason to configure the target-namespace when pipeline-resources are installed and this is addressed in a WIP PR https://github.com/openshift/tektoncd-pipeline-operator/pull/66 

The aim is to be able to provide a cluster-wide CRD like the one below 

```yaml
---
apiVersion: config.operator.tekton.dev/v1alpha1
kind: Config
metadata:
  name: cluster
spec:
  targetNamespace: tekton-piplines
```

PR #70, opened 2019-07-04
Test pr no namespace

nikhil-thomas:
Remove --namespace flag from test command
    
    Running tests in a namespace different from build namespace might be causing image pull errors
    
    This patch is to verify this hypothesis


nikhil-thomas:
obsolete, Hence closing

PR #78, opened 2019-07-16
updates oc and kubectl binaries for CI

hrishin:
Curent OC/kubectl are kind of old and doesnt
support image extraction properly

This patch updates OC/kubectl CLI


chmouel:
Can't we just use latest stable ?

chmouel:
why don't we download just latest?

hrishin:
@chmouel we can, just for the sake of [checksum](https://github.com/openshift/tektoncd-pipeline-operator/pull/78/files#diff-48abca7cd218705cd020054b38fa29cdR41) in the build we picked the particular version. 

Can we do that with latest?

chmouel:
@hrishin submmited a PR for this in #79

PR #79, opened 2019-07-16
Automatically get latest version when installing oc tools

chmouel:
/cc @hrishin


chmouel:
`latest` seems to point to `4.1.4` instead of `4.1.6` as available in the mirrors 🤷‍♂  maybe we can ping the openshift folks to check if that's a mistake?

nikhil-thomas:
@chmouel Thanks :)

hrishin:
@nikhil-thomas

PR #80, opened 2019-07-16
Update to operator-sdk 0.9.0

sthaha:
This PR introduces the following changes
- upgrades to operator-sdk version 0.9.0 which makes use of go mod
- moves from dep to go-mod
- updates vendor/
- adapts to changes to `make build`

Signed-off-by: sthaha@redhat.com


cmd/manager/main.go
```diff
@@ -25,8 +32,9 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost       = "0.0.0.0"
-	metricsPort int32 = 8383
+	metricsHost               = "0.0.0.0"
+	metricsPort         int32 = 8383
+	operatorMetricsPort int32 = 8686
```
chmouel: (Non related) Do we have plan to do something with those metrics port ?

sthaha: part of `operator-sdk` `0.9.1`

sthaha: we may, there is a plan to do operator metering.

PR #83, opened 2019-07-25
Change operator ns to openshift-operators

sthaha:
AllNamespace installation mode installs operator to `openshift-operators`,
so change deploy/*.yaml to reflect that.


nikhil-thomas:
@sthaha e2e tests are failing as the `e2e test` deployment happens in `openshift-pipelines-operators` Namespace.
Please update the `./out/test-namespace` make target in test.mk

PR #84, opened 2019-07-30
Pipelines release 0.5.2

nikhil-thomas:
Add 0.5.2 release files

Add 0.5.2 CSV file

Update operator image reference in deploy/operator.yaml and in 0.5.2 CSV

Update CatalogSource script

This patch fixes the container image build step.
The bug was occurring because the wrong binary was getting packaged
with the image build by `operator-sdk build <image:tag>` command

Add a make target `osdk-image` to ensure container image is built correctly

Modify yaml document order in deploy/resources/v0.5.2/release.yaml

Delete old pipeline payloads and old CSVs

Adds `latest` tekton cli `tkn` link in CSV description (https://github.com/tektoncd/cli/releases/latest)

Removes `operator-sdk scorecard` config .osdk-scorecard.yaml as recomended way to test CSV is using `--olm-deployed` flag
https://github.com/operator-framework/community-operators/blob/master/docs/testing-operators.md#testing-with-scorecard


scripts/olm_catalog.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 
 indent() {
-  INDENT="      "
-  sed "s/^/$INDENT/" | sed "s/^${INDENT}\($1\)/${INDENT:0:-2}- \1/"
-}
+    INDENT="      "
```
sthaha: local indent="         "

deploy/resources/v0.5.2/release.yaml
```diff
@@ -0,0 +1,769 @@
...
+  names:
+    kind: TaskRun
+    plural: taskruns
+    categories:
+      - all
+      - tekton-pipelines
+    shortNames:
+      - tr
+      - trs
+  scope: Namespaced
```
sbose78:
```
validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          description: 'APIVersion defines the versioned schema of this
            representation of an object. Servers should convert recognized
            schemas to the latest internal value, and may reject unrecognized
            values.'
          type: string
        kind:
          description: 'Kind is a string value representing the REST resource
            this object represents. Servers may infer this from the endpoint
            the client submits requests to. Cannot be updated. In CamelCase.'
          type: string
        metadata:
          type: object
        spec:
          properties:
            taskRef:
              description: taskRef refers to the Tekton Task ....... Optional.
              type: string
```

nikhil-thomas: It can be added. If we do it should be added during downstream pipelines release.
https://github.com/openshift/tektoncd-pipeline/tree/release-v0.5.2/openshift/release.

At present the admission webhook will catch wrong resource manifests.

From a user's perspective, did you run into any gotchas as `validation.openAPIV3Schema` are not there in the pipeline CRDs. It will help me strengthen the case for adding those.

scripts/olm_catalog.sh
nikhil-thomas: done

deploy/resources/v0.5.2/release.yaml
sthaha: @sbose78 agree the schema validation is useful, do you want to file an issue upstream?

nikhil-thomas: absolutely, :+1:

Issue #87, opened 2019-08-27
Automate pipeline metrics configurations from 


hrishin:
The objective of this task is to automate the following metrics related configuration thus a user can see metrics in Grafana and Prometheus as without dealing with external configuration as they install the OpenShift Pipelines.

Configurations:
* Create Prometheus scrape configuration using ServiceMonitor*
* Add RBAC for Prometheus in to perform Service discovery on pipeline deployments 

Issue: https://jira.coreos.com/browse/SRVKP-298

nikhil-thomas:
@hrishin shall we close this issue

hrishin:
@nikhil-thomas yes we can

PR #88, opened 2019-08-27
Updates reconciliation logic

hrishin:
Problem:
    - In order to get cluster-wide Prometheus to scrape the
      pipeline metrics it has to apply a certain configuration
      such as ServiceMonitor, RAB resources
    - Before applying such a configuration operator has to make sure
      pipeline controllers are up and running
    
Solution:
    - This patch adds `post-install` configuration step, which read the
      kubernets/CR manifests and applies it into target pipeline namespace
    - Changes the reconciliation flow to wait for pipeline deployments
      are up and running
    - Additionally, it adds unit tests for config API and E2E tests for
      metrics, check invlida resource(none cluster name config)resources
    
Fixes
      - https://github.com/openshift/tektoncd-pipeline-operator/issues/87


pkg/apis/operator/v1alpha1/config_types_test.go
```diff
@@ -0,0 +1,119 @@
...
+				{
+					Version: "v0.6.0",
+					Code:    ReadyStatus,
+				},
+			},
+		},
+	}
+
+	actual := config.IsUpToDateWith("v0.7.0")
+	cmp.Diff(false, actual)
```
hrishin: :man_facepalming:

pkg/controller/config/config_controller.go
```diff
@@ -225,21 +254,86 @@ func (r *ReconcileConfig) reconcileInstall(req reconcile.Request, res *op.Config
...
+	log.Info("successfully applied pipeline resources")
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileConfig) verifyDeployments(logger logr.Logger, res *op.Config) (bool, error) {
+	log.Info("verifying pipeline deployments")
+
+	deployments := appsv1.DeploymentList{}
+	listOpts := client.ListOptions{
+		Namespace: res.Spec.TargetNamespace,
```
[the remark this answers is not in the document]
hrishin: Yes that makes sense :+1:  (considering addons), previously I've added labels then dropped it back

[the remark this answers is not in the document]
nikhil-thomas: the memcacehed opeator is a good reference in this aspect
https://github.com/operator-framework/operator-sdk-samples/blob/8f06fd5435f4004fe7165cdf4091e233bdc3e764/memcached-operator/pkg/controller/memcached/memcached_controller.go#L253

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -78,3 +85,67 @@ type ConfigList struct {
 func init() {
 	SchemeBuilder.Register(&Config{}, &ConfigList{})
 }
+
+//IsUpToDateWith return true of config resource reached to its final state
+//The state could be either Success or Error one
+func (c *Config) IsUpToDateWith(version string) bool {
```
sthaha: nit: `HasVersion` may read better.

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -78,3 +85,67 @@ type ConfigList struct {
...
+
+//IsMarkedInValid return true of config resource marked as invalid
+func (c *Config) IsMarkedInValid(version string) bool {
+	cond := c.Status.Conditions
+	if len(cond) == 0 {
+		return false
+	}
+
+	latest := cond[0]
+	return latest.Code == ErrorStatus && latest.Version == version
```
sthaha: why should the Code be in `ErrorStatus` ? did you mean `!=`

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -78,3 +85,67 @@ type ConfigList struct {
...
+	if len(cond) == 0 {
+		return false
+	}
+
+	latest := cond[0]
+	return latest.Version == version && (latest.Code == ReadyStatus || latest.Code == ErrorStatus)
+}
+
+//IsMarkedInValid return true of config resource marked as invalid
+func (c *Config) IsMarkedInValid(version string) bool {
```
sthaha: How about turning this into 
1. `config.CurrentStatus` 
2. the caller does `config.CurrentStatus() != ErrorStatus` ?

pkg/apis/operator/v1alpha1/config_types.go
sthaha: Oh! I read the function name as `Valid` instead of invalid; be because of `InValid`

pkg/controller/config/config_controller.go
```diff
@@ -157,51 +168,70 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 
 	log.Info("reconciling config change")
 
-	res := &op.Config{}
```
sthaha: why is this removed?

pkg/controller/config/config_controller.go
```diff
@@ -157,51 +168,70 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
...
 		log.Info("resource has been deleted")
-		return r.reconcileDeletion(req, res)
+		return r.reconcileDeletion(req, config)
 	}
 
 	// Error reading the object - requeue the request.
 	if err != nil {
 		log.Error(err, "requeueing event since there was an error reading object")
-		return reconcile.Result{}, err
+		return reconcile.Result{Requeue: true}, err
```
sthaha: we don't have to set the `Requeue`, since `err` should trigger a reconcile IIUC.

pkg/controller/config/config_controller.go
```diff
@@ -225,21 +254,86 @@ func (r *ReconcileConfig) reconcileInstall(req reconcile.Request, res *op.Config
...
+
+	log.Info("successfully applied pipeline resources")
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileConfig) verifyDeployments(logger logr.Logger, res *op.Config) (bool, error) {
+	log.Info("verifying pipeline deployments")
+
+	deployments := appsv1.DeploymentList{}
+	listOpts := client.ListOptions{
```
sthaha: wouldn't all deployments have the owner-ref set which we should/can make use of?

deploy/post-install/2metrics-service-monitor.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: controller
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: openshift-pipeline-monitor
```
sthaha: openshift-pipeline*s*-monitor

sthaha:
@hrishin do you intend to work on this PR further? I see that the validation part of this PR is quite valuable; perhaps as a separate commit.

pkg/controller/config/config_controller.go
hrishin: It's subjective to error, the comment was saying `// Error reading the object - requeue the request.`

sthaha: Probably because https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#reconcile-loop  says so ...; hence you don't have to explicitly set Requeue.

pkg/apis/operator/v1alpha1/config_types.go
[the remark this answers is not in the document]
hrishin: Not sure, because its usage is different. One check and could reconcile further (HasVersion)  while other would not (IsMarkedInValid)

pkg/controller/config/config_controller.go
hrishin: ohh yes forget about `err`

pkg/controller/config/config_controller.go
hrishin: technically we can do it there no out of box API or declarative way of doing it

hrishin:
@sthaha will get this in ASAP

hrishin:
@sthaha @nikhil-thomas Please let me know why you think about this patch. If we think this could solve the issue of task validation. Then I can get this in ASAP.

Thanks!

hrishin:
So as per discussion, I guess @nikhil-thomas is going to fix this issue, Hence closing this PR. 
Thanks, for reviews.

Issue #89, opened 2019-09-03
apache thrift dependency is broken 


hrishin:
$subject dependency is broken and its failing the CI

```
go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /go/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:
	fatal: unable to access 'https://git.apache.org/thrift.git/': Failed connect to git.apache.org:443; Connection timed out
go: error loading module requirements
```

chmouel:
https://github.com/tektoncd/cli/commit/a0a6754f72b8f2b5845a41e2f47772e8458450d4

hrishin:
Thank you @chmouel :+1:

chmouel:
It's weird tho that cli/operator have this dep when -pipeline doesnt

hrishin:
Yes dont know, may be because of 

```
go mod graph | grep apache
go.opencensus.io@v0.18.0 git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999
github.com/openzipkin/zipkin-go@v0.1.6 github.com/eapache/queue@v1.1.0
github.com/openzipkin/zipkin-go@v0.1.6 github.com/eapache/go-xerial-snappy@v0.0.0-20180814174437-776d5712da21
github.com/openzipkin/zipkin-go@v0.1.6 github.com/eapache/go-resiliency@v1.1.0
go.opencensus.io@v0.20.0 github.com/apache/thrift@v0.12.0
go.opencensus.io@v0.19.2 github.com/apache/thrift@v0.12.0
go.opencensus.io@v0.19.1 git.apache.org/thrift.git@v0.12.0
```

vdemeester:
@chmouel @hrishin this needs to be fixed upstream too right ?

chmouel:
> @chmouel @hrishin this needs to be fixed upstream too right ?

oh i didn't realise it wasn't 👼 

as a rule we should never pr on downstream except plumbing/downstream ci fixes,

PR #90, opened 2019-09-03
fixes thrift dependency issue

hrishin:
CI build fails while vendoring the dependency for thrift.

This patch fixes the the dependency issue by replace
it with right origin url

fixes https://github.com/openshift/tektoncd-pipeline-operator/issues/89


hrishin:
Thank you folks!

PR #92, opened 2019-10-02
Auto creates SA and RBAC for pipelines

sthaha:
Previously, to run pipelines, user was asked to create a SA, configure
RBAC to grant `edit` role. Additional the SA had to be granted
`privileged` SCC. This isn't needed as granting SCC to
pipeline-controller is sufficient to start a privileged container.

This patch addresses the issue above by configuring the `privileged` SCC
and also by auto-configuring SA for all non `openshift` and `kube`
namespaces.


pkg/controller/config/config_controller.go
```diff
@@ -231,22 +238,150 @@ func (r *ReconcileConfig) reconcileInstall(req reconcile.Request, res *op.Config
...
+//log.Info("scc already in required state", "action", "none")
+//return nil
+//}
+
+//log.Info("privileged scc needs updation")
+//privileged.Users = newList
+//updated, err := r.secClient.SecurityContextConstraints().Update(privileged)
+//log.Info("updated .. scc : ", "updated", updated.Users)
+//return err
+//}
```
sthaha: should remove this

nikhil-thomas:
\retest

test/helpers/helpers.go
```diff
@@ -94,8 +119,45 @@ func DeleteClusterCR(t *testing.T, name string) {
...
+			t.Logf("failed to get privileged scc: %s \n", err)
+			return false, err
+		}
+		t.Logf("... looking at %v", privileged.Users)
+
+		ctrlSA := fmt.Sprintf("system:serviceaccount:%s:%s", ns, sa)
+		return inList(privileged.Users, ctrlSA), nil
+	})
+	AssertNoError(t, err)
+}
```
nikhil-thomas: :+1:

PR #93, opened 2019-10-02
Update readme to reflect how to override the created docker tag

brujoand:
The readme suggested that setting ´tag=yourtag´ would override the docker image tag for the created image. But this did not work, and this PR should fix that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,11 @@ cd ${GOPATH}/src/github.com/openshift/tektoncd-pipeline-operator
 ```
 2. Build go and the container image
 ```
-make osdk-image tag=${YOUR_REGISTORY}/openshift-pipelines-operator:${IMAGE_TAG}
+make osdk-image IMAGE_TAG=${YOUR_REGISTRY}/openshift-pipelines-operator:${YOUR_IMAGE_TAG}
 ```
 3. Push the container image
 ```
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -51,8 +51,10 @@ upgrade-build: #TODO: reenable it
 
 .PHONY: osdk-image
 osdk-image:
+ifndef IMAGE_TAG
+ override IMAGE_TAG = quay.io/rhpipeline/openshift-pipelines-operator:test
```
chmouel: I think you can just use : 

```
IMAGE_TAG := quay.io/rhpipeline/openshift-pipelines-operator:test
```

and having `make IMAGE_TAG=foo` would redefine it,

brujoand: ah, you are absolutely right. 👍

brujoand:
I've reverted the change to the makefile, I guess the update to the readme might still me useful.

brujoand:
/retest 
go dependency management is not my strong suit, but this seem to do the trick.

brujoand:
Awesome, I've reverted the change to ´go.mod˙ now, assuming we can just squash these commits. As for the readme, I think I've found some other things that needs to be updated but I need some help deciphering what changes are required. I've reported this as #94 if anyone as time to take a look.

PR #95, opened 2019-10-03
   Fix CI by upgrading to operator-sdk version

sthaha:
golangci-lint breaks because go list fails due to `thrift` package
    issue; see: https://github.com/golang/go/issues/34016 and later due to
    another issue https://github.com/golangci/golangci-lint/issues/658 for
    which a release of golangci-lint is yet to be made.
    This patch fixes these by moving to operator-sdk version 0.10 and by
    installing (go get) latest golangci-lint


sthaha:
> /lgtm
> @sthaha should this be done upstream too ? 👼

Thanks, that is a good point! .. shall submit one upstream.

PR #101, opened 2019-10-15
Add inject-Default-SA Transformer

nikhil-thomas:
Add an `injectDefaultSA` transfor function (manifestival.Transformer)
to inject a specific service account as the `default-service-account`
in `config-defaults` configMap in payload (release.yaml) being processed

Moves constants, flags etc into a `flags` package
Updates codebase w.r.t `flags` package path

(cherry picked from commit 8de6dccfb15ce72cc4443cc1d0b565ca7efb4e86)

### DIFF
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -2,11 +2,10 @@ package config
 
 import (
 	"context"
-	"flag"
 	"fmt"
-	"github.com/tektoncd/operator/pkg/controller/rbac"
+
+	"github.com/tektoncd/operator/pkg/controller/flags"
 	"github.com/tektoncd/operator/pkg/controller/transform"
-	"path/filepath"
 
 	"github.com/go-logr/logr"
 	mf "github.com/jcrossley3/manifestival"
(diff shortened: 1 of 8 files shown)


pkg/controller/config/config_controller.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/tektoncd/operator/pkg/controller/rbac"
```
sthaha: Could you please move the flags into a separate package so that controllers do not have to depend on each other?

pkg/controller/transform/transform.go
```diff
@@ -0,0 +1,39 @@
...
+
+		cm.Data["default-service-account"] = defaultSA
+
+		unstrObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(cm)
+		if err != nil {
+			return err
+		}
+
+		u.SetUnstructuredContent(unstrObj)
+
```
sthaha: nit: unwanted empty lines :)

pkg/controller/transform/transform.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		cm := &corev1.ConfigMap{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, cm)
+		if err != nil {
+			return err
+		}
+
+		cm.Data["default-service-account"] = defaultSA
+
```
sthaha: nit: empty line :)

pkg/controller/transform/transform.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+
+	mf "github.com/jcrossley3/manifestival"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// InjectDefaultSA adds default service account into config-defaults configMap
+func InjectDefaultSA(defaultSA string) mf.Transformer {
```
sthaha: Nice! ♥️

pkg/controller/config/config_controller.go
nikhil-thomas: done

pkg/controller/flags/flags.go
```diff
@@ -0,0 +1,64 @@
...
+	IgnorePattern   string
+	TektonVersion   = "v0.7.0"
+	ResourceWatched string
+	ResourceDir     string
+	TargetNamespace string
+	NoAutoInstall   bool
+	Recursive       bool
+)
+
+func init() {
```
sthaha: could we use a function `Parse` that gets called from `main` instead of relying on `init` magic?

pkg/controller/transform/transform.go
```diff
@@ -0,0 +1,37 @@
...
+			return nil
+		}
+
+		cm := &corev1.ConfigMap{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, cm)
+		if err != nil {
+			return err
+		}
+
+		cm.Data["default-service-account"] = defaultSA
```
sthaha: Is this string stored in a config/flag else where?

PR #105, opened 2019-11-05
OpenShift Pipelines Operator 0.8.0 Release

nikhil-thomas:
Set TektonVersion to 0.8.0

Add 0.8.0 OpenShift Pipelines payload

### DIFF
--- a/deploy/olm-catalog/openshift-pipelines-operator/0.8.0/operator_v1alpha1_config_crd.yaml
+++ b/deploy/olm-catalog/openshift-pipelines-operator/0.8.0/operator_v1alpha1_config_crd.yaml
@@ -1,0 +1,61 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: config.operator.tekton.dev
+spec:
+  group: operator.tekton.dev
+  names:
+    kind: Config
+    listKind: ConfigList
+    plural: config
+    singular: config
+  scope: Cluster
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            targetNamespace:
+              description: namespace where OpenShift pipelines will be installed
+              type: string
+          required:
+          - targetNamespace
+          type: object
+        status:
+          properties:
+            conditions:
+              description: installation status sorted in reverse chronological order
+              items:
+                properties:
+                  code:
+                    description: Code indicates the status of installation of pipeline resources.
+                    type: string
+                  details:
+                    description: Additional details about the Code
+                    type: string
+                  version:
+                    description: The version of tekton pipelines
+                    type: string
+                required:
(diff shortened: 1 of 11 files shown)


deploy/olm-catalog/openshift-pipelines-operator/0.8.0/operator_v1alpha1_config_crd.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
sthaha: why is the crd copied here? on purpose?

docs/release.md
```diff
@@ -6,9 +6,9 @@
 1. Checkout release branch
 1. Copy the release yaml from the pipelines repo to operator
    deploy/resources/<version>
-1. modify operator config_controller.go to update the <version>
+1. update operator version pkg/flag/flag.go
```
sthaha: ❤️

docs/release.md
```diff
@@ -6,9 +6,9 @@
...
 1. test the operator using `up local`
-1. build image
+1. build image (make sure that the project base diretory name is `openshift-pipelines-operator`)
```
sthaha: IIRC, it worked without the need for basedir to match image name.

pkg/flag/flag.go
```diff
@@ -26,7 +26,7 @@ const (
 var (
 	flagSet *pflag.FlagSet
 
-	TektonVersion   = "v0.7.0"
+	TektonVersion   = "v0.8.0"
```
sthaha: this should be a flag instead .. "latest" by default so that we don't have to keep changing it

deploy/olm-catalog/openshift-pipelines-operator/0.8.0/operator_v1alpha1_config_crd.yaml
nikhil-thomas: yes, we use the `--update-crds flag` so each casv version gets bundled with the right CRD revision

in addition, in operatorhub repo they want nested bundle structure. This means newer CRDs can be shipped with each csv version without breaking older versions.
```
openshift-pipelines-operator
├── 0.5.2
│   ├── config.operator.tekton.dev.crd.yaml
│   └── openshift-pipelines-operator.v0.5.2.clusterserviceversion.yaml
├── 0.7.0
│   ├── config.operator.tekton.dev.crd.yaml
│   └── openshift-pipelines-operator.v0.7.0.clusterserviceversion.yaml
├── 0.8.0
│   ├── config.operator.tekton.dev.crd.yaml
│   └── openshift-pipelines-operator.v0.8.0.clusterserviceversion.yaml
└── openshift-pipelines-operator.package.yaml
```

docs/release.md
nikhil-thomas: could you try after deleting `build/_output/bin`

pkg/flag/flag.go
nikhil-thomas: That will be an improvement. Let us add it as an improvement/tech-debt story in jira.

PR #108, opened 2019-11-15
Add buildah and s2i tasks

sthaha:
This patch adds and update-tasks.sh script which downloads tasks from
Upstream and Downstream catalog, munges them to make the tasks as
cluster-task.


scripts/update-tasks.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
+set -e -u -o pipefail
+
+declare -r SCRIPT_PATH=$(readlink -f "$0")
```
chmouel: `readlink -f` fails on *BSD (osx) :(((((

scripts/update-tasks.sh
```diff
@@ -0,0 +1,110 @@
...
+
+usage() {
+  cat <<-EOF
+    USAGE:
+      $(basename $SCRIPT_PATH) DEST_DIR
+EOF
+
+}
+
+declare -r CATALOG_VERSION="release-v0.7"
```
chmouel: can it be an argument instead of static?

deploy/resources/v0.8.0/02-clustertasks/buildah-task.yaml
```diff
@@ -0,0 +1,6 @@
+# auto generated by script/update-tasks.sh
+# DO NOT EDIT: use the script instead
+# source: https://raw.githubusercontent.com/openshift/tektoncd-catalog/release-v0.7/buildah/buildah-task.yaml
+#
+---
+
```
vdemeester: but it's empty though :angel:

deploy/resources/v0.8.0/02-clustertasks/s2i-go-task.yaml
```diff
@@ -0,0 +1,59 @@
...
+        description: Verify the TLS on the registry endpoint (for push/pull to a non-TLS registry)
+        default: "true"
+        type: string
+  outputs:
+    resources:
+      - name: image
+        type: image
+  steps:
+    - name: generate
+      image: quay.io/openshift-pipeline/s2i
```
vdemeester: Shouldn't we use tags for those ?

chmouel: Did we ever do the release of the images? I was hoping if we did,

scripts/update-tasks.sh
sthaha: thank you! I have a fix for that

deploy/resources/v0.8.0/02-clustertasks/buildah-task.yaml
sthaha: good find, the same with s2i task. I know why !.. `cat <<-EOF` does not fail :(

scripts/update-tasks.sh
sthaha: good idea ..

deploy/resources/v0.8.0/02-clustertasks/s2i-go-task.yaml
sthaha: yes, it is a very good idea to pin down versions.
An issue against catalog?

scripts/update-tasks.sh
sthaha: done

deploy/resources/v0.8.0/02-clustertasks/s2i-go-task.yaml
chmouel: We have to do quay image mirroring thing change as well to get them pinned,

deploy/resources/v0.8.0/02-clustertasks/buildah-task.yaml
```diff
@@ -0,0 +1,55 @@
+# auto generated by script/update-tasks.sh
+# DO NOT EDIT: use the script instead
+# source: https://raw.githubusercontent.com/openshift/tektoncd-catalog/release-v0.7/buildah/buildah.yaml
+#
+---
+apiVersion: tekton.dev/v1alpha1
+kind: ClusterTask
+metadata:
+  name: buildah
```
nikhil-thomas: are we planning to add version in names using the operator (buildah-v0.7.0, buildah-v0.8.0 ...)
or is it something we are going to handle in the future versions

scripts/update-tasks.sh
```diff
@@ -0,0 +1,157 @@
...
+
+  info "Downloading tasks from catalog $catalog to $dest_dir directory"
+  for t in ${tasks[@]} ; do
+    # task filenames do not follow a naming convention,
+    # some are taskname.yaml while others are taskname-task.yaml
+    # so, try both before failing
+    local task_url="$catalog/$catalog_version/$t/${t}-task.yaml"
+    local task_alt_url="$catalog/$catalog_version/$t/${t}.yaml"
+
+    local task_path="$dest_dir/$t-task.yaml"
```
nikhil-thomas: could we make it 
`    local task_path="$dest_dir/$t/$t-task.yaml"`

scripts/update-tasks.sh
```diff
@@ -0,0 +1,157 @@
...
+      die 1  "failed to convert $t to $t-$version"
+  done
+}
+
+
+create_version() {
+  local task_path="$1"; shift
+  local task="$1"; shift
+  local version="$1"; shift
+  local task_version_path="$(dirname $task_path)/$task-$version-task.yaml"
```
nikhil-thomas: and could we make this
`  local task_version_path="$(dirname $task_path)/$task/$task-$version-task.yaml"`

sthaha: that isn't needed, we just need to address the comment above to achieve the desired behaviour.

PR #109, opened 2019-11-22
Fix default tasknames

sthaha:
Ensure that task names are valid

    Tekton webhook requires that the taskname do not contain a `.`. Hence
    all clustertasks should use `-` instead of `.` and `_` can't be used
    since it fails dns 1123

    See: https://github.com/kubernetes/kubernetes/blob/d87c921a516a5b0387269910cef8551fae62de7f/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L155

    Signed-off-by: Sunil Thaha <sthah@redhat.com>


hrishin:
@sthaha lint tests for yaml are failing, I guess?

PR #112, opened 2019-11-25
Modify Dockerfile.tools

nikhil-thomas:
Recent update in dockerfile.tools (centos:7 to centos:8)
fixed git version but it ends up installing go version 1.11.6

This patch pins go version to 1.13.4 in ci.

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -13,7 +13,6 @@ RUN yum install epel-release -y
 RUN yum --enablerepo=centosplus install -y  \
     findutils \
     git \
-    golang \
     make \
     procps-ng \
     tar \
@@ -23,6 +22,11 @@ RUN yum --enablerepo=centosplus install -y  \
     yamllint \
     python3-virtualenv \
     && yum clean all
+
+RUN wget https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz; \
+    tar -C /usr/local -xzf go1.13.4.linux-amd64.tar.gz
+
+ENV PATH=$PATH:/usr/local/go/bin
 
 # install dep
 RUN mkdir -p $GOPATH/bin && chmod a+rwx $GOPATH \


openshift-ci/Dockerfile.tools
```diff
@@ -24,16 +22,13 @@ RUN yum install epel-release -y \
...
 
-# install dep
-RUN mkdir -p $GOPATH/bin && chmod a+rwx $GOPATH \
-    && curl -L -s https://github.com/golang/dep/releases/download/v0.5.1/dep-linux-amd64 -o dep \
-    && echo "7479cca72da0596bb3c23094d363ea32b7336daa5473fa785a2099be28ecd0e3  dep" > dep-linux-amd64.sha256 \
-    && sha256sum -c dep-linux-amd64.sha256 \
-    && rm dep-linux-amd64.sha256 \
-    && chmod +x ./dep \
-    && mv dep $GOPATH/bin/dep
+RUN yum -y install  https://centos7.iuscommunity.org/ius-release.rpm \
```
chmouel: what about moving to centos:8 instead?

nikhil-thomas: i didn't find one in `openshift` namespace in ci-cluster
https://github.com/openshift/release/blob/97ac15779bdfd82414b418166347018eecefe748/ci-operator/config/openshift/tektoncd-pipeline-operator/openshift-tektoncd-pipeline-operator-master.yaml#L6

i can try, but it will take more time. shall we get it working with this and move to centos:8 after the demo on wednesday

nikhil-thomas: @chmouel im pushing a patch (cnetos:8) to openshift/release

chmouel: ah yeah right you can only do from this side,

sthaha: this become much easier if we switched to centos:8

nikhil-thomas:
will have to force merge this patch as openshift/release build-root take the dockerfile.tools from the current master not from the patch

nikhil-thomas:
force merging this patch as openshift/release build-root take the dockerfile.tools from the current master not from the patch

PR #113, opened 2019-11-26
Upgrade git, dep and operator-sdk

sthaha:


openshift-ci/Dockerfile.tools
```diff
@@ -1,4 +1,4 @@
-FROM centos:7 as build-tools
+FROM centos:8 as build-tools
```
chmouel: if you use centos:8 why do you need to update git manually?

sthaha: where did I update git "manually" ?

sthaha: the `tools` image will have an updated git is what I meant

nikhil-thomas: height of best practices in git commit messages :P

nikhil-thomas:
force merging this patch as openshift/release build-root take the `dockerfile.tools` from the current master not from the patch

```
Pulling image centos:7 ...                                                                                                                        
--> FROM centos:7 as build-tools                                                                                                                  
--> LABEL maintainer="pipelines-dev <pipelines-dev@redhat.com>"   
...
```
base image should be centos:8

PR #117, opened 2019-11-27
Update README.md

bigkevmcd:
No longer using `dep` - `Gopkg.toml` has been removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,6 @@ You must install these tools:
 1. [`go`](https://golang.org/doc/install): The language Tektoncd-pipeline-operator is
    built in
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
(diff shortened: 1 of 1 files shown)


chmouel:
Why is it tho that we have `dep` in the `Dockerfile.tools` 🤔  https://github.com/openshift/tektoncd-pipeline-operator/blob/master/openshift-ci/Dockerfile.tools#L37-L44

cc @sthaha @nikhil-thomas

sthaha:
> Why is it tho that we have `dep` in the `Dockerfile.tools` 🤔 https://github.com/openshift/tektoncd-pipeline-operator/blob/master/openshift-ci/Dockerfile.tools#L37-L44

I thought of deleting it y'day when I changed the image version to centos:8 but thought I will do the cleanup in separate patch.

PR #124, opened 2019-12-03
Fix webhook timing issue while installing clustertasks and triggers

nikhil-thomas:
Modify directory v0.8.0 payload

Modify v0.8.0 payload directory structure

Rename config.operator.tekton.dev status codes


pkg/controller/validate/validate.go
```diff
@@ -0,0 +1,29 @@
+package validate
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+
```
hrishin: any reason leaving `blankspace`?

pkg/controller/config/config_controller.go
```diff
@@ -152,34 +165,46 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 		return reconcile.Result{}, err
 	}
 
-	if isUpToDate(cfg) {
+	if stateCheck(cfg, op.InstalledStatus) {
 		log.Info("skipping installation, resource already up to date")
 		return reconcile.Result{}, nil
 	}
-
+	if stateCheck(cfg, op.ValidatedPipeline) {
```
hrishin: :+1: for idiomatic status checks.

Additionally wouldn't be nice flow would be organic?
- [ ] installing pipelines if not installed
- [ ] validate pipeline if installing
- [ ] install add on if pipeline is validated

pkg/controller/config/config_controller.go
```diff
@@ -228,6 +253,101 @@ func (r *ReconcileConfig) reconcileInstall(req reconcile.Request, cfg *op.Config
...
+			Code:    op.ErrorStatus,
+			Details: err.Error(),
+			Version: flag.TektonVersion})
+		return reconcile.Result{}, err
+	}
+
+	if err := r.addons.ApplyAll(); err != nil {
+		log.Error(err, "failed to apply addons yaml manifest")
+		// ignoring failure to update
+		_ = r.updateStatus(cfg, op.ConfigCondition{
```
hrishin: Wondering if failure in stalling add-on should mark pipeline `Config` in error state?

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -45,15 +45,35 @@ type ConfigCondition struct {
...
+	InstallingAddons InstallStatus = "installing-addons"
+
+	// InstalledStatus indicates that all pipeline resources are installed successfully
+	InstalledStatus InstallStatus = "installed"
 
 	// ErrorStatus indicates that there was an error installing pipeline resources
 	// Check details field for additional details
 	ErrorStatus InstallStatus = "error"
+
+	// InstallingStatus indicates that the pipeline resources are being installed
```
sthaha: Delete?

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -45,15 +45,35 @@ type ConfigCondition struct {
...
 
-	// InstallingStatus indicates that the pipeline resources are being installed
-	InstallingStatus InstallStatus = "installing"
+	// AppliedPipeline indicates that the core pipeline resources
+	// have been applied on the cluster
+	AppliedPipeline InstallStatus = "applied-core-pipeline"
+
+	// WaitingPipelineValidation indicates that core pipeline deployments are being
+	// ensured to be running
+	WaitingPipelineValidation InstallStatus = "waiting-pipeline-validation"
```
sthaha: how about `validating-pipeline` ?

pkg/controller/config/config_controller.go
```diff
@@ -3,18 +3,19 @@ package config
...
+	"path/filepath"
+	"time"
 
 	"github.com/go-logr/logr"
 	mf "github.com/jcrossley3/manifestival"
 	sec "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	"github.com/prometheus/common/log"
 	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	"github.com/tektoncd/operator/pkg/controller/transform"
```
sthaha: could you please move non controllers code outside the `pkg/controller` ?

pkg/controller/config/config_controller.go
```diff
@@ -43,22 +44,33 @@ func init() {
 // Add creates a new Config Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	m, err := mf.NewManifest(flag.ResourceDir, flag.Recursive, mgr.GetClient())
```
sthaha: 🌟✨ 🦸‍♂️

pkg/controller/config/config_controller.go
```diff
@@ -43,22 +44,33 @@ func init() {
...
 		return err
 	}
-	return add(mgr, newReconciler(mgr, m))
+	return add(mgr, rec)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, m mf.Manifest) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
+	pplnPath := filepath.Join(flag.ResourceDir, "pipelines")
```
sthaha: nit: pipelinePath; doesn't that read better?

pkg/controller/config/config_controller.go
```diff
@@ -43,22 +44,33 @@ func init() {
...
+	ppln, err := mf.NewManifest(pplnPath, flag.Recursive, mgr.GetClient())
```
sthaha: nit: ppln -> pipeline 🙏

pkg/controller/config/config_controller.go
```diff
@@ -152,34 +165,46 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 		return reconcile.Result{}, err
 	}
 
-	if isUpToDate(cfg) {
+	if stateCheck(cfg, op.InstalledStatus) {
```
sthaha: How about calling it `isStatus(cfg. op.InstalledStatus)` also how about `cfg.IsStatus(op.InstalledStatus)` ?

pkg/controller/config/config_controller.go
sthaha: I was thinking of a state machine like this .. 

```
switch cfg.Status() {
case ErrorStatus: return applyPipeline()
case  UnknownStatus: return applyPipeline()
case  ApplyingPipeline: return applyPipeline()
case AppliedPipeline: return validatePipeline()
case ValidatingPipeline: return validatePipeline()
case ValidatedPipeline: return applyAddons()
case AppliedAddons: return markInstalled() 
case InstalledState:     return validateUptoDate() // checks uptodate ? including uuid matching
}
```

pkg/controller/config/config_controller.go
sthaha: yes, it should. These addons are mandatory and isn't managed by a different CR; so we have to indicate that failure to apply these addons as a failure.

pkg/controller/validate/validate.go
```diff
@@ -0,0 +1,29 @@
...
+		Name:      name,
+	}
+	err := c.Get(ctx, key, dp)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	if dp.Status.AvailableReplicas == *dp.Spec.Replicas {
```
sthaha: `*` is that needed?

sthaha: also `return dp.Status.AvailableReplicatas == dp.Status.Replicas , nil` removes the need for  `if`

pkg/controller/config/config_controller.go
```diff
@@ -398,16 +522,19 @@ func createCR(c client.Client) error {
 	return err
 }
 
-func isUpToDate(cfg *op.Config) bool {
+func stateCheck(cfg *op.Config, code op.InstallStatus) bool {
+	if !matchesUUID(cfg.Status.OperatorUUID) {
```
sthaha: this need not be checked for all states. If you follow the state machine above, we can move the stateCheck and the version check to just `Installed` state.

pkg/controller/config/config_controller.go
sthaha: IMHO; this state machine should suffice ... 

```
   switch cfg.InstallStatus() {
    case op.ErrorStatus:
      return r.applyPipeline(req, cfg)
    case op.EmptyStatus:
      return r.applyPipeline(req, cfg)
    case op.AppliedPipeline:
      return r.validatePipeline(req, cfg)
    case op.ValidatedPipeline:
      return r.applyAddons(req, cfg)
    case op.InstalledStatus:
      return r.validateVersion(req, cfg)
```

PR #125, opened 2019-12-03
Add pipeline samples for UI

piyush-garg:
This will add some pipeline samples which will be
shown as a reference during Create Pipeline in UI

It will be shown using the help of CRD
consoleYAMLsample and these CRs need to be applied
during the installation of pipeline using Operator

This will add CRS yaml as optional and will
include these for installation only if
consoleYAMLsample CRD is installed.


nikhil-thomas:
@piyush-garg 
could you rebasse and update the path consolesamples.
create a sub-directory inside `deploy/resources/v0.8.0`

nikhil-thomas:
looks good at present.
lets wait until we hear from @siamaksade  about how we version snippets. (one snippet per task or one snippet each for each task version).
Then we can decide how to handle @pradeepitm12 's pr https://github.com/openshift/tektoncd-pipeline-operator/pull/127 using your `conditional read of ConsoleSampleYamls mechanism` 
https://github.com/openshift/tektoncd-pipeline-operator/blob/9f90053810d2a488acac45155baa3aea4a09e1ac/pkg/controller/config/config_controller.go#L108

pkg/controller/config/config_controller.go
```diff
@@ -59,8 +59,7 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		return nil, err
 	}
 
-	addonsPath := filepath.Join(flag.ResourceDir, "addons")
-	addons, err := mf.NewManifest(addonsPath, flag.Recursive, mgr.GetClient())
+	addons, err := readAddons(mgr)
```
nikhil-thomas: can we pass `mgr.GetClient ` instead of `mgr`

piyush-garg: I need both `mgr.GetClient()` and `mgr.GetConfig()`

nikhil-thomas: the client is a dynamic client,
could you try to query CRD, like this: https://github.com/openshift/tektoncd-pipeline-operator/blob/8b7ecd3f763871596251c6191d31093adf3eb228/pkg/utils/validate/validate.go#L25

PR #138, opened 2019-12-10
Add mechanism to fetch nightly/latest pipeline payload

nikhil-thomas:
Add mechanism to fetch nightly/latest pipeline payload for nightly builds

Update update-to-head.sh to use openshift/release/fetch-pipeline.sh script

### DIFF
--- a/openshift/release/fetch-pipeline.sh
+++ b/openshift/release/fetch-pipeline.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Detect which version of pipeline should be installed
+# First it tries nightly
+# If that doesn't work it tries previous releases (until the MAX_SHIFT variable)
+# If not it exit 1
+# It can take the argument --only-stable-release to not do nightly but only detect the pipeline version
+
+# set max shift to 0, so that when a version is explicitly specified that version is fetched
+# modify this in future if a workflow based on latest version and recent (shifted) versions is needed
+MAX_SHIFT=1
+NIGHTLY_RELEASE="https://raw.githubusercontent.com/openshift/tektoncd-pipeline/release-next/openshift/release/tektoncd-pipeline-nightly.yaml"
+STABLE_RELEASE_URL='https://raw.githubusercontent.com/openshift/tektoncd-pipeline/release-${version}/openshift/release/tektoncd-pipeline-${version}.yaml'
+
+function get_version {
+    local shift=${1} # 0 is latest, increase is the version before etc...
+    local version=$(curl -s https://api.github.com/repos/tektoncd/pipeline/releases | python -c "import sys, json;x=json.load(sys.stdin);print(x[${shift}]['tag_name'])")
+    echo $(eval echo ${STABLE_RELEASE_URL})
+}
+
+function tryurl {
+    curl -s -o /dev/null -f ${1} || return 1
+}
+
+function geturl() {
+    if [[ ${1} = "release-next" ]];then
+         if tryurl ${NIGHTLY_RELEASE};then
+             echo ${NIGHTLY_RELEASE}
+             return 0
+         fi
+        for shifted in `seq 0 ${MAX_SHIFT}`;do
+            versionyaml=$(get_version ${shifted})
+            if tryurl ${versionyaml};then
+                echo ${versionyaml}
+                return 0
+            fi
+        done
+    else
+        local version=${1}
+        versionyaml=$(eval echo ${STABLE_RELEASE_URL})
+        echo ${versionyaml}
+        return 0
+    fi
+
+    echo \n"No working Pipeline payload url found"\n
+    exit 1
+}
+
+URL=$(geturl $1)
+echo Pipeline Payload URL: ${URL}
+
+[[ -d ${2}/pipelines ]] || mkdir -p ${2}/pipelines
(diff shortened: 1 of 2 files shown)


openshift/release/fetch-pipeline.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+#
+# Detect which version of pipeline should be installed
+# First it tries nightly
+# If that doesn't work it tries previous releases (until the MAX_SHIFT variable)
+# If not it exit 1
+# It can take the argument --only-stable-release to not do nightly but only detect the pipeline version
+
```
chmouel: `set -eu` may be a good idea

openshift/release/fetch-pipeline.sh
```diff
@@ -0,0 +1,58 @@
...
+    curl -s -o /dev/null -f ${1} || return 1
+}
+
+function geturl() {
+    if [[ ${1} = "release-next" ]];then
+         if tryurl ${NIGHTLY_RELEASE};then
+             echo ${NIGHTLY_RELEASE}
+             return 0
+         fi
+         echo release next logic\n
```
chmouel: you dont need a \n

openshift/release/fetch-pipeline.sh
```diff
@@ -0,0 +1,58 @@
...
+        for shifted in `seq 0 ${MAX_SHIFT}`;do
+            versionyaml=$(get_version ${shifted})
+            echo $versionyaml
```
chmouel: is that what you want to do ? echo versionyaml all the time here?

chmouel: and i think there is an issue here, this get_version doesnt install `release-next` (nightly) but a stable version.

The idea when I wrote the script on cli and catalog : 

- try install release nightly as tested on CI.
- if CI has failed  the release.yaml will not be here
- try to install the downstream stable release.
- we detect what is the latest release from upstream (as we don't have that information elsewhere)
- sometime downstream is lagging with upstream, so let's try a previous version of the release
- usually we should not be 1 version behind, if we are more than 1 we should really fail

Maybe the original script that I wrote is buggy (haven't checked) and doesn't respect that logic, if it doesn't then we would need to get it fixed.

openshift/release/fetch-pipeline.sh
nikhil-thomas: :+1: 
i think i put it there for debugging.

openshift/release/fetch-pipeline.sh
nikhil-thomas: > is that what you want to do ? echo versionyaml all the time here?

no, that was unneccessary

nikhil-thomas: > and i think there is an issue here, this get_version doesnt install `release-next` (nightly) but a stable version.
> 
> The idea when I wrote the script is :
> 
> * try install release nightly as tested on CI.
> * if CI has failed  the release.yaml will not be here
> * try to install the downstream stable release.
> * we detect what is the latest release from upstream (as we don't have that information elsewhere)
> * sometime downstream is lagging with upstream, so let's try a previous version of the release
> * usually we should not be 1 version behind, if we are more than 1 we should really fail
> 
> Maybe the original script that I wrote is buggy (haven't checked) and doesn't respect that logic, if it doesn't then we would need to get it fixed.

The `function geturl()` has the `if block` for release-next (nightly)
    - it tries to check if the nightly url works
    - if no it will find a stable url (max shift 1)

the `else` block is for releases where we need a specific stable version.

nikhil-thomas: may i should write a separate script to handle payload fetching for nightly and for releases.

openshift/release/fetch-pipeline.sh
nikhil-thomas: i missed it

openshift/release/fetch-pipeline.sh
nikhil-thomas: current implementation fetches release-next for 
```update-to-head.sh```
and stable release for
```update-to-head.sh v0.9.2```

Comment (chmouel) [reply openshift/release/fetch-pipeline.sh]: what is your definition of release ? not tektoncd release but openshift-pipeline release?

Comment (nikhil-thomas) [reply openshift/release/fetch-pipeline.sh]: openshift-pipeline release(s)
available at 
```
STABLE_RELEASE_URL='https://raw.githubusercontent.com/openshift/tektoncd-pipeline/release-${version}/openshift/release/tektoncd-pipeline-${version}.yaml'
```

as this repo has no upstream every reference is from openshift org

File: openshift/release/update-to-head.sh [hunk 4]
```diff
@@ -4,14 +4,49 @@
...
+${PROJECT_ROOT}/openshift/release/fetch-pipeline.sh ${VERSION} ${PAYLOAD_PATH}
+
+# copy rest of the payload from the previous release
+# TODO get triggers from nightly or latest release
+# TODO run scripts/update-tasks.sh to get cluster tasks
+#get triggers manifest
+#get cluster task manifest
+#get consoleSample
+LATEST_RELEASE=$(ls ${PAYLOAD_ROOT} | sort | tail -n 1)
+for d in $(ls ${PAYLOAD_ROOT}/${LATEST_RELEASE}); do
```
hrishin: what is `d` here? could we give better name?

PR #143, opened 2019-12-12
Validate the webhooks are registered

sthaha:
This patch changes the workflow as follows
1. The status after pipeline is installed is now set to `applied-pipeline`
   to avoid having to reapply the manifest if already applied successfully

2. Both deployment and mutation webhook is validate before setting the
   status to `validated-pipeline`. An additional delay of 15 seconds is
   applied before applying addons

This ensures that cluster-task defaults are set by the webhook when the
addons are applied.


pkg/controller/config/config_controller.go
```diff
@@ -292,42 +294,72 @@ func (r *ReconcileConfig) validatePipeline(req reconcile.Request, cfg *op.Config
...
+	if err != nil {
+		log.Error(err, "failed to validate mutating webhook")
+		_ = r.updateStatus(cfg, op.ConfigCondition{
+			Code:    op.PipelineValidateError,
+			Details: err.Error(),
+			Version: flag.TektonVersion})
+		return reconcile.Result{RequeueAfter: 15 * time.Second}, err
+	}
+	if !found {
+		return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
```
hrishin: What's the intent of requeing here again? :thinking:

pkg/controller/config/config_controller.go
```diff
@@ -12,9 +12,9 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	"github.com/prometheus/common/log"
 	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
-	"github.com/tektoncd/operator/pkg/controller/transform"
-	"github.com/tektoncd/operator/pkg/controller/validate"
 	"github.com/tektoncd/operator/pkg/flag"
+	"github.com/tektoncd/operator/pkg/utils/transform"
+	"github.com/tektoncd/operator/pkg/utils/validate"
```
nikhil-thomas: :+1:

pkg/controller/config/config_controller.go
```diff
@@ -292,42 +294,72 @@ func (r *ReconcileConfig) validatePipeline(req reconcile.Request, cfg *op.Config
...
-		return reconcile.Result{}, err
+		log.Error(err, "validating webhook deployment error")
+		return false, err
 	}
+
 	if !controller || !webhook {
-		log.Info("pipeline controller or/and pipeline webhook is not up")
-		return reconcile.Result{
-			Requeue:      true,
-			RequeueAfter: 5 * time.Second,
```
nikhil-thomas: @hrishin the above Reque{REqueafter: 15} is a replacement for this.
Now it is being separated for tekton-webhook and tekton-controller

pkg/controller/config/config_controller.go
nikhil-thomas: @hrishin 
https://github.com/openshift/tektoncd-pipeline-operator/pull/143#discussion_r357475469

pkg/controller/config/config_controller.go
nikhil-thomas: @hrishin https://github.com/openshift/tektoncd-pipeline-operator/pull/143#discussion_r357475469

nikhil-thomas: so if the deployment is not available yet, we need to check again, preferably a timed requeue than an a exponential requeue. So when tekton controller and tekton-webhook is available we can mark config condition `validated` and proceed with installing addons.

hrishin: Thanks  @nikhil-thomas for the explaining it :+1: . However wondering that change in state of the deployment would trigger reconciliation, thus do we need the requeue explicitely? 

Though nothing harmful as such on queueing it again, rather its guarantee that reconciliation loop would trigger again.

PR #155, opened 2019-12-19
Updates make build target for productization

hrishin:
Productization build requires vendored build and different
build output path.
This patch adds additional target for the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,7 @@ VENDOR_BUILD ?=
 
 .PHONY: build
 ## Build the operator
-build_prod: ./out/operator
-
-.PHONY: build
-## Build the operator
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,6 +15,13 @@ include ./make/test.mk
 include ./make/docker.mk
 include ./make/csv.mk
 
+BUILD_DIR ?= ./out/
+BUILD_OUT_FILE ?= operator
+VENDOR_BUILD ?= 
+
+.PHONY: build
```
sthaha: did you mean `build_prod` ?

hrishin: @sthaha I've updated the the patch. PTAL.

PR #164, opened 2019-12-24
Add mechanism to update TektonVersion flag for nightly

nikhil-thomas:
Add a `sed` step in openshift/release/fetch-pipeline.sh to
update `TektonVersion` variable in flag package.
(eg: TektonVersion="release-next")
So that the right payload is used to run tests.


piyush-garg:
lgtm

PR #170, opened 2019-12-31
Simplify nightly test payaload fetch mechanism

nikhil-thomas:
This patch is the first part of simplifying
collecting payload (pipeline, triggers, clustertasks etc)

This patch reduces the scope of

    - openshift/release/fetch-pipeline.sh
    - openshift/release/update-to-head.sh

to nightly tests payload fetching

Separate scripts, make targets, documentation will be added for
creating versioned releases (branching, payload fetching ...)

This change is necessary because the operator's dependencies are not
only pipelines but also other items like triggers, clustertasks, consoleyamlsamples ...
each with its own releases (nightly/versioned)

Hence, trying to manage both nightly builds and versioned releases for operator using the
same set of scripts lead to complications. So it is better handle them in separate scripts.

### DIFF
--- a/openshift/release/fetch-pipeline.sh
+++ b/openshift/release/fetch-pipeline.sh
@@ -25,37 +25,23 @@ function tryurl {
 }
 
 function geturl() {
-    if [[ ${1} = "release-next" ]];then
-         if tryurl ${NIGHTLY_RELEASE};then
-             echo ${NIGHTLY_RELEASE}
-             return 0
-         fi
-        for shifted in `seq 0 ${MAX_SHIFT}`;do
-            versionyaml=$(get_version ${shifted})
-            if tryurl ${versionyaml};then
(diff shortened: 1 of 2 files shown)


openshift/release/fetch-pipeline.sh
```diff
@@ -25,37 +25,23 @@ function tryurl {
...
-        done
-    else
-        local version=${1}
-        PAYLOAD_PIPELINE_VERSION=${1}
-        versionyaml=$(eval echo ${STABLE_RELEASE_URL})
-        echo ${versionyaml}
-        return 0
+    if tryurl ${NIGHTLY_RELEASE};then
+         echo ${NIGHTLY_RELEASE}
+         return 0
```
chmouel: thanks this is less confusing than before 🙇

PR #181, opened 2020-01-08
Improve Release Doc

nikhil-thomas:
Add information in installing operator-courier

Add information on making an operator version on `canary` channel
available on `dev-preview` channel


chmouel:
I would imaging the information about make csv would go to a DEVELOPMENT.md or something, and keep the FAQ user oriented? what do you think?

nikhil-thomas:
> I would imaging the information about make csv would go to a DEVELOPMENT.md or something, and keep the FAQ user oriented? what do you think?

true, we make CSVs when we need to test it against OLM and make a release. So i added to release.md. Most of the time in development we don't have to make the CSV (the simple operator.yaml deployment manifest is useful enough).

hrishin:
I agree to both. 

/lgtm

PR #187, opened 2020-01-13
Add go and java pipeline for +Add flows

pradeepitm12:
This adds pipelines for go and java based source.
Pipeline consist of build and deploy tasks
required for +Add flows in console

### DIFF
--- a/pkg/flag/flag.go
+++ b/pkg/flag/flag.go
@@ -28,6 +28,8 @@ const (
 	TriggerControllerName = "tekton-triggers-controller"
 	TriggerWebhookName    = "tekton-triggers-webhook"
 
+	AnnotationPreserveNS = "operator.tekton.dev/preserve-namespace"
+
 	uuidPath = "deploy/uuid"
 )
 
--- a/pkg/utils/transform/transform.go
+++ b/pkg/utils/transform/transform.go
@@ -35,3 +35,16 @@ func InjectDefaultSA(defaultSA string) mf.Transformer {
 		return nil
 	}
 }
+
+func InjectNamespaceConditional(preserveNamespace string, targetNamespace string) mf.Transformer {
+	tf := mf.InjectNamespace(targetNamespace)
+	return func(u *unstructured.Unstructured) error {
+		annotations := u.GetAnnotations()
+		val, ok := annotations[preserveNamespace]
+		if ok && val == "true" {
+			return nil
+		}
+
+		return tf(u)
+	}
+}
--- a/deploy/resources/v0.9.2/addons/pipelines/s2i-go/s2i-go-build-deploy.yaml
+++ b/deploy/resources/v0.9.2/addons/pipelines/s2i-go/s2i-go-build-deploy.yaml
@@ -1,0 +1,43 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: s2i-go
+  namespace: openshift
+  labels:
+    pipeline.openshift.io/runtime: golang
+  annotations:
+    operator.tekton.dev/preserve-namespace: "true"
+spec:
+  params:
+    - name: IMAGE_NAME
+      type: string
+  resources:
+    - name: app-source
+      type: git
+    - name: app-image
+      type: image
+
+  tasks:
+    - name: build
+      taskRef:
+        name: s2i-go
+        kind: ClusterTask
+      resources:
+        inputs:
+          - name: source
+            resource: app-source
+        outputs:
+          - name: image
+            resource: app-image
+      params:
+        - name: TLSVERIFY
+          value: "false"
+
+    - name: deploy
+      taskRef:
+        name: openshift-client
+        kind: ClusterTask
+      params:
+        - name: ARGS
+          value: ["new-app", "--docker-image", "$(params.IMAGE_NAME)"]
--- a/deploy/resources/v0.9.2/addons/pipelines/s2i-java-11/s2i-java-11-build-deploy.yaml
+++ b/deploy/resources/v0.9.2/addons/pipelines/s2i-java-11/s2i-java-11-build-deploy.yaml
@@ -1,0 +1,43 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: s2i-java-11
+  namespace: openshift
+  labels:
+    pipeline.openshift.io/runtime: java
+  annotations:
+    operator.tekton.dev/preserve-namespace: "true"
+spec:
+  params:
+    - name: IMAGE_NAME
+      type: string
+  resources:
+    - name: app-source
+      type: git
+    - name: app-image
+      type: image
+
+  tasks:
+    - name: build
+      taskRef:
+        name: s2i-java-11
+        kind: ClusterTask
+      resources:
+        inputs:
+          - name: source
+            resource: app-source
+        outputs:
+          - name: image
+            resource: app-image
+      params:
+        - name: TLSVERIFY
+          value: "false"
+
+    - name: deploy
+      taskRef:
+        name: openshift-client
+        kind: ClusterTask
+      params:
+        - name: ARGS
+          value: ["new-app", "--docker-image", "$(params.IMAGE_NAME)"]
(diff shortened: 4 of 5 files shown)


pradeepitm12:
@nikhil-thomas @sthaha 
This is an initial PR from a series to add pipeline template for most used languages.
Please let me know if we the directory structures and the version is handled properly.

pradeepitm12:
cc @hrishin

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
+---
```
sthaha: @pradeepitm12 why is this added to `0.8.0` ?

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - name: source
+            resource: app-source
+        outputs:
+          - name: image
+            resource: app-image
+      params:
+        - name: TLSVERIFY
+          value: "false"
+
+    - name: deploy
```
siamaksade: `runAfter` is missing in the pipelines in this PR so that deploy runs after build

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
nikhil-thomas: +1
@pradeepitm12 could you move this to `deploy/resources/v0.9.2/addons`

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: s2i-go-build-and-deploy-pipeline
+  namespace: openshift
```
nikhil-thomas: @pradeepitm12 
is the `namespace: openshift` a hard constraint for this feature to work in ui.

if no, could you remove metadata.name from the pipeline samples.

if yes, this pr will need to handle that in the controller logic.

nikhil-thomas:
@pradeepitm12 
nit: could you add a `-` in versioned names of pipeline templates

eg: change 
`s2i-gov0-8-0-build-deploy.yaml`
to
`s2i-go-v0-8-0-build-deploy.yaml`

like the clustertask versioned names: https://github.com/openshift/tektoncd-pipeline-operator/tree/master/deploy/resources/v0.9.2/addons/clustertasks/s2i-go

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
pradeepitm12: Yes, for the time being, `namespace: openshift` is a hard constraint.

nikhil-thomas:
@pradeepitm12 @vdemeester 
as per our discussion, lets add an annotation to the pipeline samples to prevent the operator from replacing `namespace: openshift` with `namespace: openshift-pipelines`

something like: `operator.tekton.dev/preserve-namespace: true`

the controller logic has to be changed to respect the annotation.

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
sthaha: 👏good catch!

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-gov0-8-0-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
+---
+apiVersion: tekton.dev/v1alpha1
```
sthaha: * could you please name the dir `pipelines` 
* and follow the same convention used in clustertasks ?
* why do we need a versioned pipeline?

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: s2i-go-build-and-deploy-pipeline
```
sthaha: should it end with `-pipeline`? isn't the `kind` already `Pipeline`?

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-java-11/s2i-java-11-build-deploy.yaml
```diff
@@ -0,0 +1,44 @@
...
+        - name: TLSVERIFY
+          value: "false"
+
+    - name: deploy
+      taskRef:
+        name: openshift-client
+        kind: ClusterTask
+      params:
+        - name: ARGS
+          value:
```
sthaha: nit: `value: [ new-app, --docker-image, $(params.IMAGE_NAME) ] ` may be easier to read

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-gov0-8-0-build-deploy.yaml
pradeepitm12: Changed to pipelines,
After testing with the console, it makes sense to have non-versioned pipeline only.

deploy/resources/v0.8.0/addons/pipelinetemplates/s2i-go/s2i-go-build-deploy.yaml
pradeepitm12: Good catch, updated.

pkg/flag/flag.go
```diff
@@ -28,6 +28,8 @@ const (
 	TriggerControllerName = "tekton-triggers-controller"
 	TriggerWebhookName    = "tekton-triggers-webhook"
 
+	PreserveNamespace = "operator.tekton.dev/preserve-namespace"
```
nikhil-thomas: nit: How about renaming the const to `PreserceNSAnnotation"

pkg/utils/transform/transform.go
```diff
@@ -35,3 +35,15 @@ func InjectDefaultSA(defaultSA string) mf.Transformer {
 		return nil
 	}
 }
+
+func InjectNamespaceCustom(preserveNamespace string, tf mf.Transformer) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
```
nikhil-thomas: as we know the `transformer` passed is always `manifetival.InjectNamespace`, we can eliminate the second argument.

before the return statement, let's add 
```
func InjectNamespaceCustom(preserveNamespace string) mf.Transformer {
    tf := mf.InjectNamespace(preserveNamespace)
    return func(u *unstructured.Unstructured) error {
    ...
    ...
```
then use tf in the returned function as it is.

pradeepitm12: Here TargetNamespace is required which comes from `cfg.Spec.TargetNamespace`

nikhil-thomas: i see, then can we pass the `targetNamespace` as the second argument to `InjectNamespaceCustom`.

and construct the namespace transformer inside InjectNamespaceCustom.
something like
```
func InjectNamespaceCustom(preserveNamespace, targetNamespace string) mf.Transformer {
    tf := mf.InjectNamespace(targetNamespace)
    return func(u *unstructured.Unstructured) error {
    ...
    ...
```

PR #192, opened 2020-01-17
removes v0-8-0 resources from v0-9-2

pradeepitm12:
resource of version v0-8-0 (specifically cluster tasks) are not
required in version 0-9-2

Signed-off-by: Pradeep Kumar pradkuma@redhat.com


nikhil-thomas:
@pradeepitm12  @vdemeester @sthaha @hrishin 
I think we should not delete old versions from new payload directories

if a user manually uninstalls (will remove all payload) and manually install a newversion, the user will not get old version clustertask. This will the defeat the purpose of providing clustertasks with versioned names.

note: at present OLM upgrade will not delete old tasks as we don't delete the set difference before applying new payload (https://github.com/openshift/tektoncd-pipeline-operator/blob/716774bf6c70f035e3e78702888a2fe37c38453b/pkg/controller/config/config_controller.go#L283)

sthaha:
@pradeepitm12 Release notes need to be updated

nikhil-thomas:
from discussion with @pradeepitm12, it turned out that current release documentation/steps is sufficient to address this change.

PR #193, opened 2020-01-17
Added runafter to sample pipelines

siamaksade:
So that sample pipelines show sequential tasks instead of independent ones.

Screenshot 2020-01-17 at 08 34 20


nikhil-thomas:
@vdemeester

PR #201, opened 2020-01-23
Add idea to git ignore

pradeepitm12:
This PR adds .idea/ to ignore GoLand IDE file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ tags
 .history
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 /tmp
+### GoLand ###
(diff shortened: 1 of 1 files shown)


nikhil-thomas:
I understand the need for this patch.
However, wouldn't it be better if we can ignore files which depends on our personal dev-env preference using a global gitignore on our local machines.

ref: [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

chmouel:
See this article too https://julien.danjou.info/properly-managing-your-gitignore/

pradeepitm12:
Thanks for the info. Closing.

PR #222, opened 2020-02-06
Add test for conditional manifest transformer

pradeepitm12:
This adds unit test for conditional manifest transformer

Signed-off-by: Pradeep Kumar pradkuma@redhat.com

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -211,3 +211,8 @@ test-upgrade-ci: ./vendor
 	$(Q)sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:tektoncd-pipeline-operator-registry-next," ./test/e2e/catalog_source_OS4.yaml | oc apply -f -
 	$(Q)./hack/check-crds-upgrade.sh
 	$(Q)operator-sdk test local ./test/e2e --no-setup --go-test-flags "-v -timeout=15m"
+
+.PHONY: test-unit
+test-unit:
+	@echo "Running unit tests..."
+	@go test -failfast -v ./pkg/...
--- a/pkg/utils/transform/transform_test.go
+++ b/pkg/utils/transform/transform_test.go
@@ -1,0 +1,54 @@
+package transform
+
+import (
+	"github.com/tektoncd/operator/pkg/flag"
+	"testing"
+
+	mf "github.com/jcrossley3/manifestival"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestTransformManifest_WithAnnotation(t *testing.T) {
+	resourceWithAnnotation := "testdata/with-annotation.yaml"
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	if err != nil {
+		t.Errorf("NewManifest() = %v, wanted no error", err)
+	}
+
+	tfs := []mf.Transformer{
+		InjectNamespaceConditional(flag.AnnotationPreserveNS, "target"),
+	}
+
+	if err := manifest.Transform(tfs...); err != nil {
+		t.Error("wanted no error :", err)
+	}
+
+	assertNamespace(t, manifest.Resources[0], "openshift")
+}
+
+func TestTransformManifest_WithoutAnnotation(t *testing.T) {
+	resourceWithoutAnnotation := "testdata/without-annotation.yaml"
+	manifest, err := mf.NewManifest(resourceWithoutAnnotation, true, nil)
+	if err != nil {
+		t.Errorf("NewManifest() = %v, wanted no error", err)
+	}
+
+	tfs := []mf.Transformer{
+		InjectNamespaceConditional(flag.AnnotationPreserveNS, "target"),
+	}
+
+	if err := manifest.Transform(tfs...); err != nil {
+		t.Error("expected no error :", err)
+	}
+
+	assertNamespace(t, manifest.Resources[0], "target")
+}
+
+func assertNamespace(t *testing.T, u unstructured.Unstructured, expected string) {
+	t.Helper()
+	v, _, _ := unstructured.NestedMap(u.Object, "metadata")
+	ns := v["namespace"]
+	if ns != expected {
+		t.Errorf("Expected '%s', got '%s'", expected, ns)
+	}
+}
--- a/pkg/utils/transform/testdata/with-annotation.yaml
+++ b/pkg/utils/transform/testdata/with-annotation.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: with-annotation
+  namespace: openshift
+  annotations:
+    operator.tekton.dev/preserve-namespace: "true"
+spec:
+  containers:
+    - name: busy
+      image: busybox
--- a/pkg/utils/transform/testdata/without-annotation.yaml
+++ b/pkg/utils/transform/testdata/without-annotation.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: with-annotation
+spec:
+  containers:
+    - name: busy
+      image: busybox


nikhil-thomas:
could you move the test into  `pkg/utils/transform` beside 
https://github.com/pradeepitm12/tektoncd-pipeline-operator/blob/addtest/pkg/utils/transform/transform.go

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+
+	mf "github.com/jcrossley3/manifestival"
+	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestTransformManifest_WithAnnotation(t *testing.T) {
+	resourceWithAnnotation := "../../../deploy/resources/v0.9.2/addons/pipelines/s2i-go/s2i-go-build-deploy.yaml"
+	assert := func(u unstructured.Unstructured, expected string) {
```
nikhil-thomas: move this assert function out of `TestTransformManifest_WithAnnotation` as it is reused in next test

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+
+	mf "github.com/jcrossley3/manifestival"
+	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestTransformManifest_WithAnnotation(t *testing.T) {
+	resourceWithAnnotation := "../../../deploy/resources/v0.9.2/addons/pipelines/s2i-go/s2i-go-build-deploy.yaml"
```
nikhil-thomas: instead of using `../../../deploy/resources/v0.9.2/addons/pipelines/s2i-go/s2i-go-build-deploy.yaml`
create a dummy file in the same location as the test or in a known test data path

the dummy file can be any kubernetes resource.

pkg/controller/config/config_controller_test.go
nikhil-thomas: name the assert function as `assertNamespace`

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	cfg := &op.Config{Spec: op.ConfigSpec{TargetNamespace: "test"}}
+
+	manifest, err := mf.NewManifest(resourceWithoutAnnotation, true, nil)
+	if err != nil {
+		t.Errorf("NewManifest() = %v, wanted no error", err)
+	}
+
+	if err := transformManifest(cfg, &manifest); err != nil {
+		t.Error("wanted no error")
```
nikhil-thomas: could we print the error as well,
something like `t.Error("wanted no error", err)`

pkg/controller/config/config_controller_test.go
nikhil-thomas: the other files can be removed or the path can change. Hence, it is better to `unit test` with test data with fixed locations

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	cfg := &op.Config{}
+
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	if err != nil {
+		t.Errorf("NewManifest() = %v, wanted no error", err)
+	}
+
+	if err := transformManifest(cfg, &manifest); err != nil {
```
nikhil-thomas: how about this: instead of using `transformManifest` from the controller.

shall we create a separate list 

```
fs := []mf.Transformer{
		transform.InjectNamespaceConditional(flag.AnnotationPreserveNS, <new namespace>),
	}
manifest.Transform(tfs...)
```

nikhil-thomas:
@pradeepitm12 we also need a make target to call the unit tests from ci.
could u add a `test-unit` target in `make/test.mk`

pkg/controller/config/config_controller_test.go
pradeepitm12: Done

make/test.mk
```diff
@@ -211,3 +211,8 @@ test-upgrade-ci: ./vendor
 	$(Q)sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:tektoncd-pipeline-operator-registry-next," ./test/e2e/catalog_source_OS4.yaml | oc apply -f -
 	$(Q)./hack/check-crds-upgrade.sh
 	$(Q)operator-sdk test local ./test/e2e --no-setup --go-test-flags "-v -timeout=15m"
+
+.PHONY: test-unit
+test-unit:
+	@echo "Running unit tests..."
+	@go test -failfast -v ./pkg/...
```
nikhil-thomas: nit: missing newline

pkg/utils/transform/transform_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if err := manifest.Transform(tfs...); err != nil {
+		t.Error("wanted no error :", err)
+	}
+
+	assertNamespace(manifest.Resources[0], "openshift", t)
+}
+
+func TestTransformManifest_WithoutAnnotation(t *testing.T) {
+	resourceWithoutAnnotation := "../../../test/testdata/resource-without-annotation.yaml"
```
nikhil-thomas: wouldn't it be better if  we keep the testdata for this unit test in `transform` folder 
` pkg/utils/transform/testdata/with-annotation.yaml`

pkg/utils/transform/transform_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if err := manifest.Transform(tfs...); err != nil {
+		t.Error("wanted no error :", err)
+	}
+
+	assertNamespace(manifest.Resources[0], "target", t)
+}
+
+func assertNamespace(u unstructured.Unstructured, expected string, t *testing.T) {
+	v, _, _ := unstructured.NestedMap(u.Object, "metadata")
```
nikhil-thomas: add a `t.Helper()` as the first line in this assert function
for better test logs

pkg/utils/transform/transform_test.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		t.Errorf("NewManifest() = %v, wanted no error", err)
+	}
+
+	tfs := []mf.Transformer{
+		InjectNamespaceConditional(flag.AnnotationPreserveNS, "target"),
+	}
+
+	if err := manifest.Transform(tfs...); err != nil {
+		t.Error("wanted no error :", err)
```
nikhil-thomas: nit: `expected` instead of `wanted`

test/testdata/resource-with-annotation.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
nikhil-thomas: nit: test data doesn't have to be a pipeline definition. Even a simple Pod definition will be fine

nikhil-thomas:
@pradeepitm12 Thank you for the updates.
I shall push a patch to openshift release to add the new unit test target to ci

Then we should see the unit tests running was part of pr ci tests

pkg/utils/transform/transform_test.go
pradeepitm12: I didn't want to put test data actual code.
But if you are fine I can add it to the suggested location.

test/testdata/resource-with-annotation.yaml
pradeepitm12: I thought of testing the actual pipeline, but it makes sense to use pod definitions for the sake of size and simplicity.

nikhil-thomas:
> @pradeepitm12 Thank you for the updates.
> I shall push a patch to openshift release to add the new unit test target to ci
> 
> Then we should see the unit tests running as part of pr ci tests

https://github.com/openshift/release/pull/7111

nikhil-thomas:
/hold
waiting for ci to detect and run unit tests

nikhil-thomas:
@pradeepitm12 and we have unit tests in ci :+1: 
/pony

PR #230, opened 2020-02-12
Add mechanism install non-Red-Hat-supported Tasks

nikhil-thomas:
Add mechanism to download and install upstream Tasks
not-supported/owned by Red Hat

Add a transform function to convert kind: Task to kind: ClusterTask

Add a mechanism to add label `operator.tekton.dev/provider-type:certified/redhat/community`.

ClusterTasks bundled with operator payload (manifest is in operator repo) are labelled
`operator.tekton.dev/provider-type:redhat`. ClusterTasks downloaded and installed by operator are labelled `operator.tekton.dev/provider-type:community`

### DIFF
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -65,18 +66,38 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		return nil, err
 	}
 
+	nonRedHat, err := fetchNonRedHat(mgr)
+	if err != nil {
+		return nil, err
+	}
+
 	secClient, err := sec.NewForConfig(mgr.GetConfig())
 	if err != nil {
 		return nil, err
 	}
 
 	return &ReconcileConfig{
-		client:    mgr.GetClient(),
-		scheme:    mgr.GetScheme(),
-		secClient: secClient,
-		pipeline:  pipeline,
-		addons:    addons,
+		client:             mgr.GetClient(),
+		scheme:             mgr.GetScheme(),
+		secClient:          secClient,
+		pipeline:           pipeline,
+		addons:             addons,
+		nonRedHatResources: nonRedHat,
 	}, nil
+}
+
+func fetchNonRedHat(mgr manager.Manager) (mf.Manifest, error) {
+	if flag.SkipNonRedHatResources {
+		return mf.Manifest{}, nil
+	}
+	//manifestival can take urls/filepaths as input
+	//more that one items can be passed as a comma separated list string
+	urls := strings.Join(flag.NonRedHatResourceURLs, ",")
+	nonRedHat, err := mf.NewManifest(urls, flag.Recursive, mgr.GetClient())
+	if err != nil {
+		return mf.Manifest{}, err
+	}
+	return nonRedHat, nil
 }
 
 // this will read all the addons files
@@ -165,11 +186,12 @@ var _ reconcile.Reconciler = &ReconcileConfig{}
 type ReconcileConfig struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client    client.Client
-	secClient *sec.SecurityV1Client
-	scheme    *runtime.Scheme
-	pipeline  mf.Manifest
-	addons    mf.Manifest
+	client             client.Client
+	secClient          *sec.SecurityV1Client
+	scheme             *runtime.Scheme
+	pipeline           mf.Manifest
+	addons             mf.Manifest
+	nonRedHatResources mf.Manifest
 }
 
 // Reconcile reads that state of the cluster for a Config object and makes changes based on the state read
@@ -215,9 +237,10 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 		return r.applyPipeline(req, cfg)
 	case op.AppliedPipeline, op.PipelineValidateError:
 		return r.validatePipeline(req, cfg)
-
 	case op.ValidatedPipeline, op.AddonsError:
 		return r.applyAddons(req, cfg)
+	case op.AppliedAddons, op.NonRedHatResourcesError:
+		return r.applyNonRedHatResources(req, cfg)
 	case op.InstalledStatus:
 		return r.validateVersion(req, cfg)
 	}
@@ -313,7 +336,47 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 			Version: flag.TektonVersion})
 		return reconcile.Result{}, err
 	}
-	log.Info("successfully applied all resources")
+	log.Info("successfully applied all addon resources")
+
+	err := r.updateStatus(cfg, op.ConfigCondition{Code: op.AppliedAddons, Version: flag.TektonVersion})
+	return reconcile.Result{Requeue: true}, err
+}
+
+func (r *ReconcileConfig) applyNonRedHatResources(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
+	log := requestLogger(req, "apply-non-redhat-resources")
+
+	// replace kind: Task, with kind: ClusterTask
+	changeKind := transform.Kind("Task", "ClusterTask")
+	if err := r.nonRedHatResources.Transform(changeKind); err != nil {
+		log.Error(err, "failed to apply manifest transformations on non Red Hat Resources")
+		// ignoring failure to update
+		_ = r.updateStatus(cfg, op.ConfigCondition{
+			Code:    op.NonRedHatResourcesError,
+			Details: err.Error(),
+			Version: flag.TektonVersion})
+		return reconcile.Result{}, err
+	}
+
+	if err := transformManifest(cfg, &r.addons); err != nil {
+		log.Error(err, "failed to apply manifest transformations on pipeline-addons")
+		// ignoring failure to update
+		_ = r.updateStatus(cfg, op.ConfigCondition{
+			Code:    op.NonRedHatResourcesError,
+			Details: err.Error(),
+			Version: flag.TektonVersion})
+		return reconcile.Result{}, err
+	}
+
+	if err := r.nonRedHatResources.ApplyAll(); err != nil {
+		log.Error(err, "failed to apply non Red Hat resources yaml manifest")
+		// ignoring failure to update
+		_ = r.updateStatus(cfg, op.ConfigCondition{
+			Code:    op.NonRedHatResourcesError,
+			Details: err.Error(),
+			Version: flag.TektonVersion})
+		return reconcile.Result{}, err
+	}
+	log.Info("successfully applied all non Red Hat resources")
 
 	err := r.updateStatus(cfg, op.ConfigCondition{Code: op.InstalledStatus, Version: flag.TektonVersion})
 	return reconcile.Result{Requeue: true}, err
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -320,10 +320,8 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 	log := requestLogger(req, "apply-addons")
 
 	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
-	addnTfrms := []mf.Transformer{
-		transform.InjectLabel(flag.LabelTaskProviderType, flag.TaskProviderTypeRedHat, true, "ClusterTask"),
-	}
-	if err := transformManifest(cfg, &r.addons, addnTfrms...); err != nil {
+	injectLabel := transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, true, "ClusterTask")
+	if err := transformManifest(cfg, &r.addons, injectLabel); err != nil {
 		log.Error(err, "failed to apply manifest transformations on pipeline-addons")
 		// ignoring failure to update
 		_ = r.updateStatus(cfg, op.ConfigCondition{
@@ -355,7 +353,7 @@ func (r *ReconcileConfig) applyNonRedHatResources(req reconcile.Request, cfg *op
 	addnTfrms := []mf.Transformer{
 		// replace kind: Task, with kind: ClusterTask
 		transform.ReplaceKind("Task", "ClusterTask"),
-		transform.InjectLabel(flag.LabelTaskProviderType, flag.TaskProviderTypeCommunity, true),
+		transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeCommunity, true),
 	}
 	if err := transformManifest(cfg, &r.nonRedHatResources, addnTfrms...); err != nil {
 		log.Error(err, "failed to apply manifest transformations on pipeline-addons")
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -320,7 +320,7 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 	log := requestLogger(req, "apply-addons")
 
 	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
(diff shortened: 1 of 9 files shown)


pkg/controller/config/config_controller.go
```diff
@@ -65,20 +67,41 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
...
+		addons:             addons,
+		nonRedHatResources: nonRedHat,
 	}, nil
 }
 
+func fetchNonRedHat(mgr manager.Manager) (mf.Manifest, error) {
+	if flag.SkipNonRedHatResources {
+		return mf.Manifest{}, nil
+	}
+	//manifestival can take urls/filepaths as input
```
sthaha: did you apply go fmt?

pkg/controller/config/config_controller.go
```diff
@@ -294,7 +319,11 @@ func matchesUUID(target string) bool {
 func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
 	log := requestLogger(req, "apply-addons")
 
-	if err := transformManifest(cfg, &r.addons); err != nil {
+	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
+	addnTfrms := []mf.Transformer{
```
sthaha: no need of an array here ..

pkg/flag/flag.go
```diff
@@ -28,23 +28,34 @@ const (
 	TriggerControllerName = "tekton-triggers-controller"
 	TriggerWebhookName    = "tekton-triggers-webhook"
 
-	AnnotationPreserveNS = "operator.tekton.dev/preserve-namespace"
+	AnnotationPreserveNS      = "operator.tekton.dev/preserve-namespace"
+	LabelTaskProviderType     = "operator.tekton.dev/provider-type"
```
sthaha: `ProviderTypeLabel` may read better.

pkg/flag/flag.go
```diff
@@ -28,23 +28,34 @@ const (
...
+	TaskProviderTypeCommunity = "community"
```
sthaha: `RedHatProvider` may read better since it can be applied to various reosources.

pkg/flag/flag.go
```diff
@@ -28,23 +28,34 @@ const (
...
+	PipelineSA             string
+	IgnorePattern          string
+	ResourceWatched        string
+	ResourceDir            string
+	TargetNamespace        string
+	NoAutoInstall          bool
+	SkipNonRedHatResources bool
+	Recursive              bool
+	OperatorUUID           string
+	NonRedHatResourceURLs  = []string{
```
sthaha: how about using a config-map for this?

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+func ReplaceKind(fromKind, toKind string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := u.GetKind()
+
+		if kind != fromKind {
+			return nil
+		}
+		err := unstructured.SetNestedField(u.Object, toKind, "kind")
+		if err != nil {
+			return fmt.Errorf("could change resource KIND, %q", err)
```
sthaha: typo: .. how about `failed to convert resource ..`

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+		}
+		return nil
+	}
+}
+
+func InjectLabel(key, value string, overwrite bool, kinds ...string) mf.Transformer {
```
sthaha: can this be unit-tested?

sthaha: may want to return an error if `kinds` length is 0

sthaha: or even better have `kind string, rest ...string` so that user is forced to pass a kind atleast.

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+	return func(u *unstructured.Unstructured) error {
+		kind := u.GetKind()
+		if len(kinds) != 0 && !itemInSlice(kind, kinds) {
```
sthaha: can avoid this check if the `InjectLabel` itself validates that kinds is non-empty.

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+			return nil
+		}
+		labels, found, err := unstructured.NestedMap(u.Object, "metadata", "labels")
+		if err != nil {
+			return fmt.Errorf("could not find labels set, %q", err)
+		}
+		if !overwrite && found {
```
sthaha: an use && here instead of inner if

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+	}
+}
+
+func InjectLabel(key, value string, overwrite bool, kinds ...string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := u.GetKind()
+		if len(kinds) != 0 && !itemInSlice(kind, kinds) {
+			return nil
+		}
+		labels, found, err := unstructured.NestedMap(u.Object, "metadata", "labels")
```
sthaha: NestedStringMap - right?

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+		if err != nil {
+			return fmt.Errorf("could not find labels set, %q", err)
+		}
+		if !overwrite && found {
+			if _, ok := labels[key]; ok {
+				return nil
+			}
+		}
+		if !found || labels == nil {
```
sthaha: I think `!found` is sufficient  or just `labels == nil`

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+			labels = map[string]interface{}{}
+		}
+		labels[key] = value
+		err = unstructured.SetNestedMap(u.Object, labels, "metadata", "labels")
```
sthaha: nit: `if err := ... ; err != nil { `

pkg/utils/transform/transform.go
```diff
@@ -48,3 +49,54 @@ func InjectNamespaceConditional(preserveNamespace string, targetNamespace string
...
+		if err != nil {
+			return fmt.Errorf("error updateing labes for %v:%v, %v", kind, u.GetName(), err)
```
sthaha: typo: 
 - updating
 - labels

sthaha: all %v can be %s

pkg/controller/config/config_controller.go
```diff
@@ -313,18 +342,53 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
...
 	}
-	log.Info("successfully applied all resources")
+	log.Info("successfully applied all addon resources")
+
+	err := r.updateStatus(cfg, op.ConfigCondition{Code: op.AppliedAddons, Version: flag.TektonVersion})
+	return reconcile.Result{Requeue: true}, err
+}
+
+func (r *ReconcileConfig) applyNonRedHatResources(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
+	log := requestLogger(req, "apply-non-redhat-resources")
```
sthaha: how about: `apply-community-resources`

pkg/controller/config/config_controller.go
```diff
@@ -313,18 +342,53 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
...
+
+	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
+	addnTfrms := []mf.Transformer{
+		// replace kind: Task, with kind: ClusterTask
+		transform.ReplaceKind("Task", "ClusterTask"),
+		transform.InjectLabel(flag.LabelTaskProviderType, flag.TaskProviderTypeCommunity, true),
```
sthaha: true is a bit misleading .. it is better to use an enum `overwrite | retain `

pkg/controller/config/config_controller.go
```diff
@@ -313,18 +342,53 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
...
 		return reconcile.Result{}, err
 	}
-	log.Info("successfully applied all resources")
+	log.Info("successfully applied all addon resources")
+
+	err := r.updateStatus(cfg, op.ConfigCondition{Code: op.AppliedAddons, Version: flag.TektonVersion})
+	return reconcile.Result{Requeue: true}, err
+}
+
+func (r *ReconcileConfig) applyNonRedHatResources(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
```
hrishin: Applying the community tasks appears bit different than applying Red Hat `ClusterTasks`. In a sense RH Tasks are applied from directory and this one applies from URLs. Any particular reason to fragment the method of applying tasks in this way?

pkg/controller/config/config_controller.go
nikhil-thomas: yes, it is configured to run automatically.
This looks like github counting tabs differently as we don't have a `.editorconfig` file.

pkg/flag/flag.go
nikhil-thomas: individually yes,
but when the labels and annotations are grouped in adjacent lines, the `<TYPE><NAME>` looks better.

In addition, it follows custom error definitions is go. `ERROR<TYPE/MESSAGE>`

pkg/utils/transform/transform.go
nikhil-thomas: That will be better. :+1:

pkg/utils/transform/transform.go
nikhil-thomas: yes, i am adding tests.

by keeping `kind` variadic, if we don't pass any kind the transformer will add labels to all resources.

pkg/utils/transform/transform.go
nikhil-thomas: the idea is to allow len(kinds) to be 0 and add lables to any kind

for example, in pipeline-core manifests or addons manifest

pkg/utils/transform/transform.go
nikhil-thomas: true, that will be better

pkg/utils/transform/transform.go
nikhil-thomas: true

pkg/controller/config/config_controller.go
nikhil-thomas: this issue explains the difference https://issues.redhat.com/browse/SRVKP-548

pkg/flag/flag.go
nikhil-thomas: im bit confused here. Are you suggesting that he key should be `RedHatProvider`, `CommunityProvider`, `CertifiedProvider`

pkg/flag/flag.go
nikhil-thomas: we can technically do it, but it will change the order of manifest creation.
First the config-map will have to be created as part of `addons`.
Then, in the next phase we will have to read the configMap, download tasks and apply them.

In the current scope, as we are not providing reconcile capability if the contents of configMap changes, IMHO i think it will be an overkill.

@sthaha @hrishin

pkg/utils/transform/transform.go
nikhil-thomas: my reasoning is, `if len(kinds)==0` that should mean `any` resource
I will add better comments

pkg/utils/transform/transform.go
nikhil-thomas: the if statement becomes too long.
so keeping it as it is.

pkg/controller/config/config_controller.go
nikhil-thomas: done

pkg/utils/transform/transform.go
nikhil-thomas:
```
//InjectLabel adds label key:value to a resource
// overwritePolicy decides whehter to overwite an already existing label
// []kinds specify the Kinds on which the label should be applied
// if len(kinds) = 0, label will be apllied to all/any resources irrespective of its Kind
```

nikhil-thomas:
@sthaha @hrishin i have pushed a patch with changes based on review comments.
I have also added unit tests
```
--- PASS: TestReplaceKind (0.00s)
    --- PASS: TestReplaceKind/should_replace_Kind_when_resource_kind_==_fromKind (0.00s)
    --- PASS: TestReplaceKind/should_not_replace_Kind_when_resource_kind_!=_fromKind (0.00s)

--- PASS: TestInjectLabel (0.00s)
    --- PASS: TestInjectLabel/should_add_label_to_a_resource (0.00s)
    --- PASS: TestInjectLabel/should_add_label_if_kind(s)_is_specified_and_does_not_match_resource_kind (0.00s)
    --- PASS: TestInjectLabel/should_retain_original_label_with_overwritePolicy_'Retain' (0.00s)
    --- PASS: TestInjectLabel/should_overwrite_original_label_with_overwritePolicy_'Overwrite' (0.00s)
    --- PASS: TestInjectLabel/should_add_labels_only_to_specified_resources (0.00s)
```

pkg/controller/config/config_controller.go
```diff
@@ -76,9 +84,24 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		secClient: secClient,
 		pipeline:  pipeline,
 		addons:    addons,
+		community: community,
 	}, nil
 }
 
+func fetchCommuntiyResources(mgr manager.Manager) (mf.Manifest, error) {
```
hrishin: Would rather like to write `func (mgr *manager.Manager) fetchCommuntiyResources(mf.Manifest, error)`?

nikhil-thomas: done

pkg/controller/config/config_controller.go
```diff
@@ -54,17 +55,24 @@ func Add(mgr manager.Manager) error {
...
 func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
+	log := ctrlLog.WithName("new-reconciler")
 	pipelinePath := filepath.Join(flag.ResourceDir, "pipelines")
 	pipeline, err := mf.NewManifest(pipelinePath, flag.Recursive, mgr.GetClient())
 	if err != nil {
 		return nil, err
 	}
 
-	addons, err := readAddons(mgr)
+	addons, err := readAddons(&mgr)
```
sthaha: Why is mgr passed as * ? isn't manager.Manager is an interface - right?

pkg/controller/config/config_controller.go
```diff
@@ -76,14 +84,29 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
 		secClient: secClient,
 		pipeline:  pipeline,
 		addons:    addons,
+		community: community,
 	}, nil
 }
 
+func fetchCommuntiyResources(mgr *manager.Manager) (mf.Manifest, error) {
```
sthaha: 😕 pointer to an interface? 🤔

sthaha: why not just accept the client if all this func needs is a client from Manager?

pkg/flag/flag.go
```diff
@@ -28,23 +28,34 @@ const (
...
 
-	TektonVersion   = "v0.9.2"
-	PipelineSA      string
-	IgnorePattern   string
-	ResourceWatched string
-	ResourceDir     string
-	TargetNamespace string
-	NoAutoInstall   bool
-	Recursive       bool
-	OperatorUUID    string
```
vdemeester: Shouldn't this be v0.10.x ?

nikhil-thomas: it is `v0.10.x` on release-branch,
this feature is being merged to master.

In fact, it should be a placeholder like `devel`. At present the version in master branch is `0.9.2`

nikhil-thomas: While I was typing this message, the 0.10.x release branch got merged.
Now you are right, it should be 0.10.1
Rebasing.

PR #245, opened 2020-02-25
Fix all pipeline templates

pradeepitm12:
Pipeline `deploy` step fails as it tries to deploy new application
deployment config is already created by `console +Add`

Instead deploy step should `rollout` the existing deployment


andrewballantyne:
@pradeepitm12 We are pivoting away from `APPLICATION_NAME` and going with the pre-existing `APP_NAME` that we have in the UI for ease of implementation / support of this feature. @siamaksade asked me to log https://issues.redhat.com/browse/SRVKP-631

pradeepitm12:
@andrewballantyne Thanks for the heads up. I will update the PR accordingly.

PR #251, opened 2020-03-02
Allows overirding the images for all pipelines components 

hrishin:
To test the productization work flow,  system point to OpenShift internal registry URL for operator and other pipelines components. All manifests for pipelines are containing in the operator image. In release process, operator CSV gets released to stage environment and same CSV gets published to prod environment. Through out this process CSV can be changed after staging environment which points to the operator image. In staging environment image URLS are different and prod environment image URL are different, hence the hard-coded images URL in all manifest will not work. 

In order to address this issue, this patch adds images overriding mechanism so through release process images cloud be configured as per environment.

Fixes https://issues.redhat.com/browse/SRVKP-635


pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+		_, err = r.applyAddons(req, config)
+
+		// THEN
+		assetNoEror(err, "failed to reconcile for applyPipeline;", t)
+		assertContainerHasImage(deployment, container, image, r.client, t)
+		assertContainerArgHasImage(deployment, arg, argImage, r.client, t)
+	})
+}
+
+func configFor(name string, namespace string) *op.Config {
```
sthaha: how about newConfig?

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: op.ConfigSpec{
+			TargetNamespace: namespace,
+		},
+	}
+}
+
+func requestFor(configName string, namespace string) reconcile.Request {
```
sthaha: how about `newRequest(..) *reconcile.Request `

README.md
```diff
@@ -3,6 +3,22 @@
...
+# Configurations
+
+### Override Images:
+Following environment variables allows overriding images of pipelines, trigger and 
+`Task` addons components
+
+1) `IMAGE_PIPELINE_<deployment-image-name>` e.g. `IMAGE_PIPELINE_WEBHOOK`
+It allows overriding pipelines or triggers deployment images. Note, `*_PIPELINE_*` 
+will override images in pipelines manifest only. Same way `*_TRIGGERS_*`
+2) `IMAGE_PIPELINE_<deployment-image-arg-name>` e.g. `IMAGE_PIPELINE_-NOP`
```
sthaha: is `IMAGE_PIPELINE_-NOP` a valid shell variable?

pkg/controller/config/config_controller.go
```diff
@@ -251,7 +251,7 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 func (r *ReconcileConfig) applyPipeline(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
 	log := requestLogger(req, "apply-pipeline")
 
-	if err := transformManifest(cfg, &r.pipeline); err != nil {
+	if err := transformManifest(cfg, &r.pipeline, transform.ReplaceDeploymentImages("IMAGE_PIPELINE_")); err != nil {
```
sthaha: `transform.DeploymentImage` may read better 👼

pkg/controller/config/config_controller_test.go
sthaha: as a rule of thump all function that uses `For` should be of the form `xForY` e.g. `priceForApple`

pkg/utils/transform/transform.go
```diff
@@ -109,6 +112,136 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+		name := strings.ToLower(container.Name)
+		if url, exist := images[name]; exist {
+			containers[i].Image = url
+		}
+	}
+}
+
+func replaceContainersArgsImage(containers []corev1.Container, images map[string]string) {
+	for c, container := range containers {
+		if len(container.Args) == 0 {
```
sthaha: if (len) isn't needed as `for` below will not execute

README.md
hrishin: right, felt so as well, shall convert to `_ARG_` or assume replacement should look for image `name` or `arg`?

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+	return reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      configName,
+			Namespace: namespace,
+		},
+	}
+}
+
+func assetNoEror(err error, msg string, t *testing.T) {
```
nikhil-thomas: typo: `assert`

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,206 @@
...
+	if containers == nil && len(containers) == 0 {
+		t.Fatalf("assertion failed; deployment not containers")
+	}
+
+	return containers
+}
+
+func mfFor(resource string, cl client.Client) (mf.Manifest, error) {
+	_, filename, _, _ := rt.Caller(0)
+	root := path.Join(path.Dir(filename), "../../..")
```
nikhil-thomas: should we keep some test data in ` pkg/controller/config/` instead of using the `deploy/resources/...`

README.md
```diff
@@ -3,6 +3,22 @@
...
+It allows overriding pipelines or triggers deployment images of containers `args`. 
+Note, `**_PIPELINE_**` will override images in pipelines manifest only. Same way `**_TRIGGERS_**`  
+3) `IMAGE_ADDONS_<PARAM>_<NAME>` e.g. `IMAGE_ADDONS_PARAM_BUILDER_IMAGE` 
+It allows overriding `ClusterTask` addon images. the `_PARAMA_` will replace the
+value of `Task.Spec.Patam,` without it will replace step image.
```
nikhil-thomas: is it `_PARAMA_` or `_PARAM_` or `_PARAM_A`

typo: `Task.Spec.Patam,`

pkg/controller/config/config_controller.go
nikhil-thomas: shall we move `IMAGE_PIPELINE_` prefix string into a `config` in `pkg/flag/flag.go`

pkg/controller/config/config_controller.go
```diff
@@ -321,7 +321,9 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 
 	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
 	injectLabel := transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, transform.Overwrite, "ClusterTask")
-	if err := transformManifest(cfg, &r.addons, injectLabel); err != nil {
+	replaceTriggerImage := transform.ReplaceDeploymentImages("IMAGE_TRIGGERS_")
+	replaceTaskImages := transform.ReplaceTaskImage("IMAGE_ADDONS_", "PARAM")
+	if err := transformManifest(cfg, &r.addons, injectLabel, replaceTriggerImage, replaceTaskImages); err != nil {
```
nikhil-thomas: nit: 
how about 
```
tfs := []mf.Transformer{
	injectLabel := transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, transform.Overwrite, "ClusterTask"),
	replaceTriggerImage := transform.ReplaceDeploymentImages("IMAGE_TRIGGERS_"),
	replaceTaskImages := transform.ReplaceTaskImage("IMAGE_ADDONS_", "PARAM"),
}
	if err := transformManifest(cfg, &r.addons, tfs...); err != nil {


```

pkg/utils/transform/transform.go
```diff
@@ -109,6 +112,136 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+	images := map[string]string{}
+	for _, env := range os.Environ() {
+		keyValue := strings.Split(env, "=")
+		key := strings.ToLower(keyValue[0])
+		prefix = strings.ToLower(prefix)
+		if !strings.HasPrefix(key, prefix) {
+			continue
+		}
+
+		name := strings.TrimPrefix(key, prefix)
```
nikhil-thomas: For the images map (and while naming ENV vars) should we use the Resource Name as well in the name. 
`name := Resource.Name + strings.TrimPrefix(key, prefix)`

Will there be a scenario where 2 Deployments have the same `arg name` and the `key` gets overwritten

I have same concern for ClusterTask - Step Images.

nikhil-thomas: may be this kind of collision is handled already, ignore this comment if that is the case

pkg/utils/transform/transform_test.go
```diff
@@ -55,6 +59,13 @@ func TestReplaceKind(t *testing.T) {
 	})
 }
 
+func assertKind(t *testing.T, u unstructured.Unstructured, kind string) {
+	t.Helper()
+	if k := u.GetKind(); k != kind {
+		t.Errorf("expected kind %s, got kind %s", kind, k)
+	}
+}
+
```
nikhil-thomas: nit: move this assertFunc to the end of the file with other assertFunc

pkg/utils/transform/transform_test.go
```diff
@@ -171,3 +168,172 @@ func assertOnResourceList(t *testing.T, items []unstructured.Unstructured, key,
...
+		deployment := deploymentFor(t, resource)
+		containers := deployment.Spec.Template.Spec.Containers
+
+		for _, container := range containers {
+			if len(container.Args) == 0 {
+				continue
+			}
+
+			for a, argument := range container.Args {
+				if argument == arg && container.Args[a+1] != image {
```
nikhil-thomas: why is it `[a+1]` instead of `[a]`

pkg/utils/transform/transform_test.go
```diff
@@ -171,3 +168,172 @@ func assertOnResourceList(t *testing.T, items []unstructured.Unstructured, key,
...
+		err = manifest.Transform(transform.ReplaceDeploymentImages("IMAGE_PIPELINE_"))
+
+		assetNoEror(t, err)
+		assertDeployContainerArgsHasImage(t, manifest.Resources, "-bash", image)
+		assertDeployContainerArgsHasImage(t, manifest.Resources, "-git", "git")
+	})
+
+	t.Run("of_task_addons", func(t *testing.T) {
+		image := "foo.bar/image/buildah"
+		os.Setenv("IMAGE_ADDONS_PARAM_BUILDER_IMAGE", image)
```
nikhil-thomas: will it be better if we name our variables as
`IMAGE_ADDONS_CLUSTERTASK_<NAME>_STEP_NAME`

pkg/utils/transform/transform_test.go
```diff
@@ -171,3 +168,172 @@ func assertOnResourceList(t *testing.T, items []unstructured.Unstructured, key,
...
+		assetNoEror(t, err)
+		err = manifest.Transform(transform.ReplaceTaskImage("IMAGE_ADDONS_", "PARAM_"))
+
+		assetNoEror(t, err)
+		assertTaskImage(t, manifest.Resources, "push", image)
+		assertParamHasImage(t, manifest.Resources, "BUILDER_IMAGE", image)
+	})
+}
+
+func assetNoEror(t *testing.T, err error) {
```
nikhil-thomas: typo: `assert`

nikhil-thomas: may be we can move all the assert functions in to a separate file in `pkg/utils/<something like test helpers>`

nikhil-thomas:
@hrishin 
```
pkg/controller/config/config_controller_test.go:192:63: Error return value of `(github.com/openshift/client-go/security/clientset/versioned/typed/security/v1.SecurityContextConstraintsInterface).Create` is not checked (errcheck)
```

pkg/utils/transform/transform.go
hrishin: >Will there be a scenario where 2 Deployments have the same arg name and the key gets overwritten

true, It could happen but not the case right now. However, if we its critical then we can address it.

pkg/utils/transform/transform_test.go
hrishin: ahh, its testing the URL of the image.

e.g. args:["-bash", "quay.io/pipeline/bash"], so when it assert for for bash -> quay.io/pipeline/bash, so ~ args[a] -> args[a+1].

right?

pkg/utils/transform/transform_test.go
hrishin: we can. The only reason I moved those function close to the caller because often time its easy find it.

pkg/utils/transform/transform.go
nikhil-thomas: sounds good

pkg/utils/transform/transform_test.go
nikhil-thomas: i see.

hrishin:
/hold

So thing is `applyAddOn` is failing. It appears that while doing `ClusterTask` transformation the way it reads `spec.steps` is introducing the unwanted side effect.

I gues it coud be an issue with [Step ](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/task_types.go#L106)API. I deally it should have `json:",inline"` annotation

hrishin:
This should fix the addon transformation https://github.com/tektoncd/pipeline/issues/2150 issue

README.md
```diff
@@ -3,6 +3,29 @@
...
+2) `IMAGE_PIPELINE__ARG_<DEPLOYMENT-IMAGE-ARG-NAME>` e.g. `IMAGE_PIPELINE_ARG_NOP`
+It allows overriding pipelines or triggers deployment images of containers `args`. 
+Note, `*_PIPELINE_ARG_*` will override images in pipelines manifest only. Same way `**_TRIGGERS_ARG_**`  
+3) `IMAGE_ADDONS_<STEP-NAME>` e.g. `IMAGE_ADDONS_PUSH` 
+It allows overriding `ClusterTask` addons steps images.
+4) `IMAGE_ADDONS_PARAM_<NAME>` e.g. `IMAGE_ADDONS_PARAM_BUILDER` 
+It allows overriding `ClusterTask` addons params images. the `_PARAM_` will replace the
+value of `Task.Spec.Param`.
+
+`Note: BE CAUTIOUS WHILE SUBSTITUTING THE IMAGES. FOR INSTANCE, IF DEPLOYMENT HAS MORE THAN ONE CONTAINER AND
```
nikhil-thomas: :+1:

pkg/controller/config/config_controller.go
```diff
@@ -251,7 +252,8 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
 func (r *ReconcileConfig) applyPipeline(req reconcile.Request, cfg *op.Config) (reconcile.Result, error) {
 	log := requestLogger(req, "apply-pipeline")
 
-	if err := transformManifest(cfg, &r.pipeline); err != nil {
+	images := toLowerCaseKeys(imagesFromEnv("IMAGE_PIPELINE_"))
```
nikhil-thomas: can we move 
`toLowerCaseKeys(imagesFromEnv("IMAGE_PIPELINE_"))`
into the transfor function `transform.DeploymentImages`
and pass `prefix string` as an arg to the `transform.DeploymentImages`

pkg/controller/config/config_controller.go
```diff
@@ -320,8 +322,14 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 	log := requestLogger(req, "apply-addons")
 
 	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
-	injectLabel := transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, transform.Overwrite, "ClusterTask")
-	if err := transformManifest(cfg, &r.addons, injectLabel); err != nil {
+	triggerImages := toLowerCaseKeys(imagesFromEnv("IMAGE_TRIGGERS_"))
+	addonImages := toLowerCaseKeys(imagesFromEnv("IMAGE_ADDONS_"))
```
nikhil-thomas: same comments for these 2
https://github.com/openshift/tektoncd-pipeline-operator/pull/251/files#r388725616

pkg/controller/config/config_controller.go
```diff
@@ -346,6 +354,17 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 	return reconcile.Result{Requeue: true}, err
 }
 
+func toLowerCaseKeys(keyValues map[string]string) map[string]string {
```
nikhil-thomas: may be we can move this to `pkg/utis`

pkg/controller/config/config_controller.go
```diff
@@ -638,3 +657,19 @@ func requestLogger(req reconcile.Request, context string) logr.Logger {
 		"Request.NamespaceName", req.NamespacedName,
 		"Request.Name", req.Name)
 }
+
+func imagesFromEnv(prefix string) map[string]string {
```
nikhil-thomas: shall we move this also to utils, especially if we do
https://github.com/openshift/tektoncd-pipeline-operator/pull/251/files#r388725616

pkg/controller/config/config_controller_test.go
```diff
@@ -0,0 +1,209 @@
...
+		)
+
+		// GIVEN
+		os.Setenv("IMAGE_PIPELINE_"+strings.ToUpper(container), image)
+		os.Setenv("IMAGE_PIPELINE_ARG_"+strings.ToUpper(arg), argImage)
+		config := newConfig(configName, namespace)
+		cl := feedConfigMock(config)
+		secCl := feedSCCMock("privileged", t)
+		pipelines, err := mfFor("pipelines", cl)
+		assetNoEror(err, "failed to create manifestival for pipelines;", t)
```
nikhil-thomas: asse`r`t :)

pkg/controller/config/config_controller_test.go
nikhil-thomas: i think it got in back again

pkg/utils/transform/transform.go
```diff
@@ -109,6 +113,98 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+		replaceContainersArgsImage(&container, images)
+	}
+}
+
+func replaceContainersArgsImage(container *corev1.Container, images map[string]string) {
+	for a, arg := range container.Args {
+		argument := strings.TrimPrefix(arg, "-")
+		argument = ARG_PREFIX + strings.ToLower(argument)
+		if url, exist := images[argument]; exist {
+			container.Args[a+1] = url
```
nikhil-thomas: just thinking out loud,
what if i write the arg list as 
`-arg=value` instead of `-arg value`

hrishin: Wise thought I'm hearing it. I will add the patch for it. I felt we could also file an issue on pipeline, triggers to accept it as `ENV VAR` or other way for accepting the configuration.

README.md
```diff
@@ -3,6 +3,31 @@
 ## FAQ
 [click here](/docs/faq.md)
 
+# Configurations
+
+### Override Images:
+Following environment variables allows overriding images of pipelines, trigger and 
+`Task` addons components
```
nikhil-thomas: nit: why is it `Task` while pipelines and trigger are in lowercase :P

pkg/controller/config/config_controller.go
```diff
@@ -320,8 +322,14 @@ func (r *ReconcileConfig) applyAddons(req reconcile.Request, cfg *op.Config) (re
 	log := requestLogger(req, "apply-addons")
 
 	//add TaskProviderType label to ClusterTasks (community, redhat, certified)
-	injectLabel := transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, transform.Overwrite, "ClusterTask")
-	if err := transformManifest(cfg, &r.addons, injectLabel); err != nil {
+	triggerImages := transform.ToLowerCaseKeys(imagesFromEnv("IMAGE_TRIGGERS_"))
+	addonImages := transform.ToLowerCaseKeys(imagesFromEnv("IMAGE_ADDONS_"))
+	addnTfrms := []mf.Transformer{
+		transform.InjectLabel(flag.LabelProviderType, flag.ProviderTypeRedHat, transform.Overwrite, "ClusterTask"),
+		transform.DeploymentImages(triggerImages),
```
nikhil-thomas: can we pass prefixes like these `"IMAGE_TRIGGERS_"`  to transformer and 
call transform.ToLowerCaseKeys(imagesFromEnv("IMAGE_TRIGGERS_")) from the transformer factory function

pkg/utils/transform/transform_test.go
```diff
@@ -109,14 +117,112 @@ func TestInjectLabel(t *testing.T) {
...
+		manifest, err := mf.NewManifest(testData, true, nil)
+		assertNoEror(t, err)
+		err = manifest.Transform(DeploymentImages(images))
+
+		assertNoEror(t, err)
+		assertDeployContainerArgsHasImage(t, manifest.Resources, "-bash", image)
+		assertDeployContainerArgsHasImage(t, manifest.Resources, "-git", "git")
+	})
+
+	t.Run("of_container_args_has_equal", func(t *testing.T) {
```
nikhil-thomas: :+1

pkg/utils/transform/transform.go
```diff
@@ -109,6 +118,151 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+			containers[i].Image = url
+		}
+
+		replaceContainersArgsImage(&container, images)
+	}
+}
+
+func replaceContainersArgsImage(container *corev1.Container, images map[string]string) {
+	for a, arg := range container.Args {
+		if argVal, hasArg := splitsByEqual(arg); hasArg {
```
nikhil-thomas: :+1:

pkg/utils/transform/transform_test.go
hrishin: Its better to address in the another PR https://github.com/openshift/tektoncd-pipeline-operator/issues/263

pkg/utils/transform/transform.go
```diff
@@ -109,6 +118,151 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+	}
+
+}
+
+func formKey(prefix, arg string) string {
+	argument := strings.ToLower(arg)
+	if prefix != "" {
+		argument = prefix + "_" + argument
+	}
+	argument = strings.ReplaceAll(argument, "-", "_")
```
sthaha: nit: return strings.ReplaceAll(...)

pkg/utils/transform/transform.go
```diff
@@ -109,6 +118,151 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+	return argument
+}
+
+func splitsByEqual(arg string) ([]string, bool) {
+	values := strings.Split(arg, "=")
+	if len(values) == 2 {
```
sthaha: can't this be > 2?

pkg/utils/transform/transform.go
```diff
@@ -109,6 +118,151 @@ func InjectLabel(key, value string, overwritePolicy OverwritePolicy, kinds ...st
...
+		name, ok := param["name"].(string)
+		if !ok {
+			transformLog.Info(fmt.Sprintf("Unable to get the step name %v", p))
+			continue
+		}
+
+		name = formKey(PARAM_PREFIX, name)
+		image, found := override[name]
+		if !found || image == "" {
+			transformLog.Info(fmt.Sprintf("Image not found for %s param, skipping replacement", name))
```
sthaha: it is better to use multiple fields / structured logging here e.g. `transformLog.Info("Image not found", "param", name, "action",  "skip")`

pkg/utils/transform/transform.go
```diff
@@ -117,3 +271,14 @@ func ItemInSlice(item string, items []string) bool {
 	}
 	return false
 }
+
+func ToLowerCaseKeys(keyValues map[string]string) map[string]string {
```
sthaha: should this be exported?

pkg/controller/config/config_controller.go
hrishin: sounds true, initially it was like that only. Later we decoupled it to use the map because of two reasons
1) So the transformation could be more tested by making map dependency, that we could inject it during the test
2) Source could be anything may be ENV,  some CR values. It would be easy to reuse while porting it for the upstream use case

pkg/utils/transform/transform.go
hrishin: Right though not to address such case. Usually users may not pass such arguments?

pkg/utils/transform/transform.go
hrishin: yes

pkg/controller/config/config_controller.go
hrishin: Let it be close to this code since its not being used anywhere else.

PR #281, opened 2020-04-08
Fix amending SCC privileged that blocks cluster upgrade

sthaha:
Previously, operator used to append `tekton-pipeline-controller` SA to
    the SCC `privileged` which prevented openshift cluster upgrade[doc].
    From the [doc]:

    > Do not modify the default SCCs. Customizing the default SCCs can
    > lead to  issues when OpenShift Container Platform is upgraded.
    > Instead, create new SCCs.

    The solution is to create a new SCC like privileged which this patch implements.

    [doc]: https://docs.openshift.com/container-platform/4.3/authentication/managing-security-context-constraints.html

### DIFF
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -483,73 +483,45 @@ func (r *ReconcileConfig) serviceAccountNameForDeployment(deployment types.Names
 
 func (r *ReconcileConfig) addPrivilegedSCC(sa string) error {
 	log := ctrlLog.WithName("scc").WithName("add")
-	privileged, err := r.secClient.SecurityV1().SecurityContextConstraints().Get("privileged", metav1.GetOptions{})
+	// get the scc and not nothing if it exists
+	scc := r.secClient.SecurityV1().SecurityContextConstraints()
+	sccPriv, err := scc.Get(flag.OperatorPrivilegedSCC, metav1.GetOptions{})
+	if err == nil {
+		log.Info("scc already exists")
+		return nil
+	} else if !errors.IsNotFound(err) {
+		log.Error(err, "failed to get scc")
+		return err
+	}
+
+	// create new copied for privileged
+	privileged, err := scc.Get("privileged", metav1.GetOptions{})
 	if err != nil {
 		log.Error(err, "scc privileged get error")
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/config/config_controller.go
```diff
@@ -483,73 +483,44 @@ func (r *ReconcileConfig) serviceAccountNameForDeployment(deployment types.Names
...
-	privileged, err := r.secClient.SecurityV1().SecurityContextConstraints().Get("privileged", metav1.GetOptions{})
-	if err != nil {
-		log.Error(err, "scc privileged get error")
+	// get the scc and not nothing if it exists
+	scc := r.secClient.SecurityV1().SecurityContextConstraints()
+	sccPriv, err := scc.Get(flag.OperatorPrivilegedSCC, metav1.GetOptions{})
+	if err == nil || !errors.IsNotFound(err) {
+		log.Error(err, "failed to find scc")
 		return err
 	}
```
alexeykazakov, line 492:
```suggestion
	// get the scc and do nothing if it exists
	scc := r.secClient.SecurityV1().SecurityContextConstraints()
	sccPriv, err := scc.Get(flag.OperatorPrivilegedSCC, metav1.GetOptions{})
	if err == nil {
		log.Info("scc already exist")
		return nil
	} else if !errors.IsNotFound(err) {
		log.Error(err, "failed to find scc")
		return err
	}
```

pkg/controller/config/config_controller.go
```diff
@@ -483,73 +483,44 @@ func (r *ReconcileConfig) serviceAccountNameForDeployment(deployment types.Names
...
+func (r *ReconcileConfig) removePrivilegedSCC() error {
+	log := ctrlLog.WithName("scc").WithName("remove")
+	scc := r.secClient.SecurityV1().SecurityContextConstraints()
 
-func addToList(list []string, item string) ([]string, bool) {
-	for _, v := range list {
-		if v == item {
-			return list, false
-		}
+	if err := scc.Delete(flag.OperatorPrivilegedSCC, &metav1.DeleteOptions{}); err != nil {
```
nikhil-thomas, line 517: nice, now we can just delete, if present. instead of having to do a remove item and patch

PR #300, opened 2020-04-28
Update e2e test target

nikhil-thomas:
Refactor lint make target

Remove hack as the root issue is fixed in golangci-lint

Update golangci-lint to v1.25.1

Add e2e-tests.sh script to improve handling of new ci IMAGE_FORMAT variable substitution.


nikhil-thomas:
@sthaha

PR #301, opened 2020-04-28
Fix config CRD spec for OCP 4.5

nikhil-thomas:


nikhil-thomas:
@sthaha @vdemeester

PR #308, opened 2020-05-04
Update non redhat resources to use image value from env

savitaashture:
JIRA : [SRVKP-697](https://issues.redhat.com/browse/SRVKP-697)

**Changes:**
[Kn](https://github.com/openshift/tektoncd-pipeline-operator/blob/master/pkg/flag/flag.go#L57) has been installed as part of [applyCommunityResources](https://github.com/openshift/tektoncd-pipeline-operator/blob/master/pkg/controller/config/config_controller.go#L357)
Added logic to update the params image

**Test:**
set
```
export IMAGE_ADDONS_PARAM_KN_IMAGE=registry.redhat.io/openshift-serverless-1/client-kn-rhel8
```
and run 
```
operator-sdk up local --namespace=""
```
to verify the `kn` task updated with the provided image

To make sure the updated image works fine test with `kn` **[TaskRun](https://github.com/openshift/tektoncd-catalog/tree/release-v0.11/kn)**

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -24,6 +24,8 @@ spec:
         env:
         - name: WATCH_NAMESPACE
           value: ""
+        - name: IMAGE_ADDONS_PARAM_KN_IMAGE
+          value: registry.redhat.io/openshift-serverless-1/client-kn-rhel8
         - name: POD_NAME
           valueFrom:
             fieldRef:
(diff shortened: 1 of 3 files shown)


savitaashture:
> Should we use the upstream image or using our own image ? (I guess we do productize one ?)

Yes we need to use productized `kn` image 
There is already a task [here](https://github.com/openshift/tektoncd-pipeline-operator/blob/master/pkg/flag/flag.go#L57)  which install `kn` so may be i need to update this PR based on that 
Testing changes with Productized `kn` image

PR #325, opened 2020-05-29
Fix e2e

piyush-garg:
Remove uused argument to pipelinesetup function
and use the right flag in ValidateSCCRemoved
function

### DIFF
--- a/test/testsuites/auto_install_pipeline.go
+++ b/test/testsuites/auto_install_pipeline.go
@@ -61,5 +61,5 @@ func ValidateDeletion(t *testing.T) {
 		flag.TriggerControllerName,
 		flag.TriggerWebhookName)
 
-	helpers.ValidateSCCRemoved(t, cr.Spec.TargetNamespace, flag.PipelineControllerName)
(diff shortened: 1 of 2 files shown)


test/testsuites/auto_install_pipeline.go
```diff
@@ -61,5 +61,5 @@ func ValidateDeletion(t *testing.T) {
 		flag.TriggerControllerName,
 		flag.TriggerWebhookName)
 
-	helpers.ValidateSCCRemoved(t, cr.Spec.TargetNamespace, flag.PipelineControllerName)
+	helpers.ValidateSCCRemoved(t, cr.Spec.TargetNamespace, flag.PipelineControllerSA)
```
nikhil-thomas, line 64: why is this change necessary, as both are the same.
https://github.com/piyush-garg/tektoncd-pipeline-operator/blob/10bd9e3ba51ec87c1a99a42dd99ee626d04daa9e/pkg/flag/flag.go#L20
```
PipelineControllerName = "tekton-pipelines-controller"
PipelineControllerSA   = "tekton-pipelines-controller"
```
in this context `PipelineControllerName` is appropriate

PR #330, opened 2020-06-05
fix crashloopbackoff of tekton-pipelines-webhook

pradeepitm12:
with pipeline 0.12.1 tekton-pipelines-webhook pod goes to
crashloopbackoff state as it tries to get pipelines webhook
service in different namespace

this adds transform function to put the correct namespace

### DIFF
--- a/pkg/utils/transform/transform.go
+++ b/pkg/utils/transform/transform.go
@@ -65,8 +65,45 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
 		if ok && val == "true" {
 			return nil
 		}
-
 		return tf(u)
+	}
+}
+
+func InjectNamespaceRoleBindingSubjects(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := strings.ToLower(u.GetKind())
+		if kind != "rolebinding" {
+			return nil
+		}
+		subjects, found, err := unstructured.NestedFieldNoCopy(u.Object, "subjects")
+		if !found || err != nil {
+			return err
+		}
+		for _, subject := range subjects.([]interface{}) {
+			m := subject.(map[string]interface{})
+			if _, ok := m["namespace"]; ok {
+				m["namespace"] = targetNamespace
+			}
+		}
+		return nil
+	}
+}
+
+func InjectNamespaceCRDWebhookClientConfig(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := strings.ToLower(u.GetKind())
+		if kind != "customresourcedefinition" {
+			return nil
+		}
+		service, found, err := unstructured.NestedFieldNoCopy(u.Object, "spec", "conversion", "webhookClientConfig", "service")
+		if !found || err != nil {
+			return err
+		}
+		m := service.(map[string]interface{})
+		if _, ok := m["namespace"]; ok {
+			m["namespace"] = targetNamespace
+		}
+		return nil
 	}
 }
 
--- a/pkg/utils/transform/transform_test.go
+++ b/pkg/utils/transform/transform_test.go
@@ -398,3 +398,99 @@ func assertTaskImage(t *testing.T, resources []unstructured.Unstructured, name s
 		}
 	}
 }
+
+func TestTransformManifest_InjectNamespaceRoleBindingSubjects(t *testing.T) {
+	resourceWithAnnotation := "testdata/inject-ns-rolebinding.yaml"
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	assertNoEror(t, err)
+	tf := InjectNamespaceRoleBindingSubjects("target")
+	err = manifest.Transform(tf)
+	assertNoEror(t, err)
+	assertRBSubjectNamespace(t, manifest.Resources[0], "target")
+}
+
+func TestTransformManifest_InjectNamespaceRoleBindingSubjects_NoNs(t *testing.T) {
+	resourceWithAnnotation := "testdata/inject-ns-rolebinding-no-ns.yaml"
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	assertNoEror(t, err)
+	tf := InjectNamespaceRoleBindingSubjects("target")
+	err = manifest.Transform(tf)
+	assertNoEror(t, err)
+	assertRBSubjectNoNamespace(t, manifest.Resources[0])
+}
+
+func TestTransformManifest_InjectNamespaceCRDWebhookClientConf(t *testing.T) {
+	resourceWithAnnotation := "testdata/inject-ns-crd-webhookclientconf.yaml"
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	assertNoEror(t, err)
+	tf := InjectNamespaceCRDWebhookClientConfig("target")
+	err = manifest.Transform(tf)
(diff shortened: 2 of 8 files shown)


nikhil-thomas:
@pradeepitm12 thank you for the fix :+1: 

Could you add unit tests for the 2 new transformers

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,44 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
 		if ok && val == "true" {
 			return nil
 		}
-
 		return tf(u)
 	}
 }
 
+func InjectNamespaceRoleBinding(targetNamespace string) mf.Transformer {
```
nikhil-thomas, line 72: could you rename it to indicate that it changes the namespace in subjects

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,44 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
+	return func(u *unstructured.Unstructured) error {
+		switch strings.ToLower(u.GetKind()) {
+		case "rolebinding":
```
sthaha, line 75: shouldn't we handle clusterrolebindings as well?

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,44 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
 			return nil
 		}
-
 		return tf(u)
 	}
 }
 
+func InjectNamespaceRoleBinding(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		switch strings.ToLower(u.GetKind()) {
```
sthaha, line 74: nit: this can be better written as 

```go

return func (u ...) error {
  kind := strings.ToLower(u.GetKind());   
  if kind != "rolebindings" {
    return nil  
  }
subjects, _, _ := unstructured.NestedField...
return nil 
}
```

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,44 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
+					m["namespace"] = targetNamespace
+				}
+			}
+			return nil
+		}
+		return nil
+	}
+}
+
+func InjectNamespaceWebhookClientConfig(targetNamespace string) mf.Transformer {
```
nikhil-thomas, line 89: could you rename this function to indicate that we are editing a CRD

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,44 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
+	return func(u *unstructured.Unstructured) error {
+		switch strings.ToLower(u.GetKind()) {
+		case "customresourcedefinition":
+			service, _, _ := unstructured.NestedFieldNoCopy(u.Object, "spec", "conversion", "webhookClientConfig", "service")
+			if service != nil {
```
sthaha, line 94: this can be made flat by 
```
  if service == nil {
     return nil  
  }
```

but shouldn't you make use of the `found` return value ? 

```go

service, found, err := unstructured.NestedFieldNoCopy(...)
if !found || err != nil {
  return nil ??
}
```

pkg/utils/transform/transform.go
nikhil-thomas: @sthaha `ClusterRoleBindings` subjects are handled by manifestival, but it doesn't care for `rolebindings` 
they should have their reasons

pkg/utils/transform/transform.go
pradeepitm12: Yes, it makes sense, I have updated both transform functions accordingly. Thanks

pkg/utils/transform/transform.go
```diff
@@ -65,11 +65,48 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
+	}
+}
+
+func InjectNamespaceCRDWebhookClientConfig(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		kind := strings.ToLower(u.GetKind())
+		if kind != "customresourcedefinition" {
+			return nil
+		}
+		service, found, err := unstructured.NestedFieldNoCopy(u.Object, "spec", "conversion", "webhookClientConfig", "service")
```
nikhil-thomas, line 98: nit, can we write is as 
```
service, found, err := unstructured.NestedFieldNoCopy(u.Object, "webhookClientConfig", "service")
```
or
```
service, found, err := unstructured.NestedFieldNoCopy(u.Object, "conversion", "webhookClientConfig", "service")```

File: pkg/utils/transform/transform_test.go [hunk 7]
```diff
@@ -398,3 +398,73 @@ func assertTaskImage(t *testing.T, resources []unstructured.Unstructured, name s
...
+}
+
+func TestTransformManifest_InjectNamespaceCRDWebhookClientConf_NoNs(t *testing.T) {
+	resourceWithAnnotation := "testdata/inject-ns-crd-webhookclientconf-no-ns.yaml"
+	manifest, err := mf.NewManifest(resourceWithAnnotation, true, nil)
+	assertNoEror(t, err)
+	tf := InjectNamespaceCRDWebhookClientConfig("target")
+	err = manifest.Transform(tf)
+	assertNoEror(t, err)
+	assertCRDWebhookClientConfNS(t, manifest.Resources[0], "target")
```
nikhil-thomas, line 439: @pradeepitm12  the last parameter should be `""`. 
the tests are passing because we have the `if _, ok := m["namespace"]; ok {` check in the assertions.

it will be better if you could split both assertions (crd and rolebinding) into 2 (total 4)
- assertRBSubjectNamespace
- assertRBSubjectNoNamespace
- assertCRDWebhookClientConfNS
- assertCRDWebhookClientConfNoNS

it will be more readable. It took me a minute and a try with golang repl to understand why `assertCRDWebhookClientConfNS(t, manifest.Resources[0], "target")` passed. :) 

nit: let us stick to `Namespace` or `NS` (assertRBSubjectNamespace and assertCRDWebhookClientConfNS)

pkg/utils/transform/transform.go
pradeepitm12: as the service available in `spec.conversion.webhookClientConfig.service` we will need to fetch spec and then put that as unstructured.spec and can do `conversion.webhookClientConfig.service` 
please do let me know if we can do something else with this.

pkg/utils/transform/transform_test.go
nikhil-thomas: i appreciate you adding the negative tests as well :+1:

pradeepitm12: I actually wrote the positive test and then added the negative one while adding the negative I added `if _, ok := m["namespace"]; ok {`.
But it actually makes sense to split both the assertions.
Thank you.

pkg/utils/transform/transform.go
nikhil-thomas: sure, please try this. 
we don't have to specify the entire path from the root.


in this pr itself, in rolebinding transformer we didn't have to use `"spec", "subjects"` 

i believe the path just has to be `distinct enough` but now absolute.
if you think using full path is more readabe, then lets use that.
```
subjects, found, err := unstructured.NestedFieldNoCopy(u.Object, "subjects")
```

PR #331, opened 2020-06-17
WIP: experiment to see what happens when priv scc binding removed

gabemontero:
rather than trying to set it all up locally, I'll see what PR e2e testing uncovers ;-)

I'll then see about deploying / testing locally like the test job does 

@nikhil-thomas @sthaha @siamaksade


gabemontero:
Interesting @nikhil-thomas @sthaha @siamaksade @vdemeester @chmouel 

at least with the e2e's in this repo deleting the priv-scc role binding in the devel and 0.11.3 subfolder (presumably the latest versions?) had no adverse effect.

Should more versions be updated?

Do I need to make these changes in other repo's that have requisite e2e's?

Or is manual testing required?

thanks

chmouel:
do we actually have e2e test testing the `openshift-pipelines-privileged` feature @sthaha  ?

sthaha:
@chmouel , @gabemontero Currently we do not have e2e tests in operator that checks if a Pipeline/Task runs fine after the controller is installed let alone the tasks that requires `privileged`  which explains why the tests didn't fail as all it validates is that controllers come up fine. 

I do agree having these tests are more than good to have.

gabemontero:
> @chmouel , @gabemontero Currently we do not have e2e tests in operator that checks if a Pipeline/Task runs fine after the controller is installed let alone the tasks that requires `privileged` which explains why the tests didn't fail as all it validates is that controllers come up fine.
> 
> I do agree having these tests are more than good to have.
> 
> Created: #332

OK, I think this is what I'm going to do
1) I'll make an attempt at going through the flow at https://github.com/openshift/tektoncd-pipeline-operator#install-olm with the priv scc binding removed, and see if / where it breaks down, and if it breaks down, sort out what finer grained, precise permissions are needed

2) run the e2e tests out of https://github.com/openshift/tektoncd-pipeline/blob/master/Makefile#L20 against a cluster from 1) and see what if anything breaks

As I mentioned in the f2f, I'm assuming buildah related tests are going to fail, cause running buildah non-privileged is *VERY* non-trivial .... we've got a separate build api / build v2 thread going on that 

The backup plan for that would be that rather than add priv scc to the controller, we should add priv scc, or whatever escalated privs are needed, to the local SA, like the `pipeline` SA ... that is one element that has already materialized as a "best practice" out of the auth design I've got going upstream.

Thoughts @sthaha @nikhil-thomas @chmouel @vdemeester ?

gabemontero:
> > @chmouel , @gabemontero Currently we do not have e2e tests in operator that checks if a Pipeline/Task runs fine after the controller is installed let alone the tasks that requires `privileged` which explains why the tests didn't fail as all it validates is that controllers come up fine.
> > I do agree having these tests are more than good to have.
> > Created: #332
> 
> OK, I think this is what I'm going to do
> 
>     1. I'll make an attempt at going through the flow at https://github.com/openshift/tektoncd-pipeline-operator#install-olm with the priv scc binding removed, and see if / where it breaks down, and if it breaks down, sort out what finer grained, precise permissions are needed

OK minikube without docker stinks :-(

@sthaha @nikhil-thomas is there a recipe for testing out OLM and the operator from this repo using CRC or openshift ?

> 
>     2. run the e2e tests out of https://github.com/openshift/tektoncd-pipeline/blob/master/Makefile#L20 against a cluster from 1) and see what if anything breaks
> 
> 
> As I mentioned in the f2f, I'm assuming buildah related tests are going to fail, cause running buildah non-privileged is _VERY_ non-trivial .... we've got a separate build api / build v2 thread going on that
> 
> The backup plan for that would be that rather than add priv scc to the controller, we should add priv scc, or whatever escalated privs are needed, to the local SA, like the `pipeline` SA ... that is one element that has already materialized as a "best practice" out of the auth design I've got going upstream.
> 
> Thoughts @sthaha @nikhil-thomas @chmouel @vdemeester ?

sthaha:
@gabemontero we use CRC for development and to test the operator, this is what I do

1. install & start crc
1. oc login ...
1. `oc apply -f deploy/crds/operator_v1alpha1_config_crd.yaml    ` 
1. operator-sdk run --local --namespace "" --operator-flags '--zap-encoder=console --recursive' 2>&1 | tee tmp/operator.log  

which applies the crd and runs the operator locally as the logged in user (this also allows you to login as the serviceaccount of the operator and test).

Hope that helps. 


Here is how I build and run operator whenever a file changes 👼
```
 fd .go$ cmd pkg | 
  entr -rc -s \
  'operator-sdk up local --namespace "" --operator-flags "--zap-encoder=console --recursive" 2>&1 | ag -v ctrl.rbac | tee tmp/operator.log'
```

@nikhil-thomas did spend some time tweaking the makefiles but I am not used to it.

sthaha:
> As I mentioned in the f2f, I'm assuming buildah related tests are going to fail, cause running buildah non-privileged is VERY non-trivial .... we've got a separate build api / build v2 thread going on that

All tasks that have `privileged: true` should fail IIUC, which includes `buildah`, `s2i` `buildpacks/buildpacks-v3.yaml`  in the upstream catalog see e.g. https://github.com/tektoncd/catalog/blob/4ce31b6c8d5f5d14c9c59735d23c8c25924dc0fd/buildah/buildah.yaml#L43


> The backup plan for that would be that rather than add priv scc to the controller, we should add priv scc, or whatever escalated privs are needed, to the local SA, like the pipeline SA ... that is one element that has already materialized as a "best practice" out of the auth design I've got going upstream.

I was thinking we could provide cluster admin to allow `scc` to be disable (thereby preserving the current behaviour) by adding a flag to the `config` crd so taht the spec may look like 

```yaml

spec:
  sccPrivileges: disable
```

This would disable SCC privilege given to the controller and still allow for the `pipeline` SA that operator creates in all namespaces to be given  scc `privileged` by a cluster-admin if the choose to .

gabemontero:
thanks for the additional info here @sthaha on both how to dev and the fact that is was not just buildah, but s2i/buildpacks as well

and the follow ups in slack @nikhil-thomas 

I hope to give it a try later today.

I *suspect* we could solve the need for priv with s2i in similar fashion as we will with buildah, but we'll see.  Less sure about buildpacks, but we might be able to broaden the effort on the buildv2 side to deal with that.

On the `sccPrivileges` notion in the config CRD, I definitely think that is a step in the right direction.  At the moment, I at least would view it as a "back up option", in that it will be much safer to manage this via updated privs on the SA used for the Tasks, etc.   

If the controller never has privs, that eliminates the escalation path entirely.

I do understand that is a broader and more labor intensive effort (both upstream and downstream), and that will have bearing on the path ultimately chosen.

gabemontero:
OK got some results
- using this branch, I was able to install pipelines from this PRs branch into my personal openshift AWS cluster using `operator-sdk` 
- verified the priv scc rolebinding deleted via this PR was not present
- ran `oc adm policy who-can use scc privileged ` and confirmed the SA `tekton-pipelines-controller` in the `openshift-pipelines` namespace is *NOT* listed
- the operator was able to populate `openshift-pipelines` with pipeline/trigger/webhook pods, no problem (perhaps you all were sure of this, but I did not want to assume this would work without priv scc)
- I did see this warning in the logs but am unsure if it is related to the removal of priv scc:  `{"level":"warn","logger":"tekton.taskrun-controller","caller":"taskrun/taskrun.go:154","msg":"Failed to log the metrics : pod has never got scheduled","knative.dev/controller":"taskrun-controller"}
`
- was able to create a new namespace and I see the `pipeline` SA (again, did not want to assume this would occur without priv scc)

So yes it seems like we are down to specific tasks/taskrun type issues.  Good to know / confirm (though again, maybe you all knew this already).

- for my buildah example, I went with https://github.com/openshift/pipelines-catalog/pull/27 cause I needed to revisit this anyway and make sure it still worked :-) ... to mix things up, I switched to the `pipelines` SA
- initially, it failed as expected with the controller not having priv scc: 
```
'failed to create task run pod "imagestream-with-buildah": pods "imagestream-with-buildah-pod-m7bmc"
        is forbidden: unable to validate against any security context constraint:
        [spec.containers[2].securityContext.privileged: Invalid value: true: Privileged
        containers are not allowed spec.containers[3].securityContext.privileged:
        Invalid value: true: Privileged containers are not allowed]. Maybe missing
        or invalid Task ggmtest/imagestream-with-buildah'
```
- I then ran `oc adm policy add-scc-to-user privileged -z pipelines`
- `oc adm policy who-can use scc privileged ` confirms it took
- **INTERESTINGLY I GET THE SAME ERROR WHEN I RECREATE THE TASKRUN**

So I've got some tekton controller / openshift apiserver debug to embark on.

gabemontero:
ahhh ... had a typo ... `pipeline` vs. `pipelines` :-)

and I just destroyed my cluster :-(

redoing ...

gabemontero:
OK, confirmed, with the right SA name, i.e. 

`oc adm policy add-scc-to-user privileged -z pipeline`

I can successfully create my priv build task pod even when the controller does not have priv scc:

```
$ oc get taskrun
NAME                       SUCCEEDED   REASON      STARTTIME   COMPLETIONTIME
imagestream-with-buildah   True        Succeeded   2m37s       19s
```

So we are now @nikhil-thomas @sthaha @vdemeester @chmouel at deciding about 
- is priv given to the controller SA or on a per namespace SA (either default or pipeline) ... per namespace SA is more secure, but has working out of the box implications (i.e. updates to catalog samples)
- is default the old behavior of priv scc enabled, or not  (preserving existing behavior, albeit pre GA vs. more secure)

I see the config api type in this repo.

What repo/code drives the UI for the operator in the console?

thanks
-

nikhil-thomas:
Hi Gabe,
              So far I was focused on getting it to work. It is very
interesting to flow your findings 👍.

> What repo/code drives the UI for the operator in the console?
openshift/console:
https://github.com/openshift/console/tree/master/frontend/packages/operator-lifecycle-manager

*info:* the controller creates the `pipeline` service account in the name
spaces which *does not* have *openshift-* prefix.

*another gotcha*
You might be aware of this already, if you are running the operator with
`operator-sdk run --local` command the operator controller is run with
`kubeadmin` privileges. When the operator runs on the cluster, it is run
using the deployment manifest, RBAC and Service Account in the deploy
folder (
https://github.com/openshift/tektoncd-pipeline-operator/tree/master/deploy)

*you might find the following make targets useful.*
opo-image - build image

opo-operator-yaml-update - updates the new image in the deployment manifest

Thank you

*Nikhil Thomas*
*Senior Software Engineer, Devtools*

nikthoma@redhat.com <dffrench@redhat.com>

Red Hat

Bengaluru, India


<https://www.redhat.com/>

gabemontero:
Thanks for the info @nikhil-thomas 

Some follow ups:

1) on 

openshift/console:
https://github.com/openshift/console/tree/master/frontend/packages/operator-lifecycle-manager

I don't see any openshift-pipelines specific stuff there.  Hence, probably cannot have any special UI around "which SAs can be privileged" etc.  Which I think is a point @sthaha made recently as well in one of the calls.

does my understanding line up with you two?

2) the kubeadmin and `operator-sdk run --local` point was *NOT* apparent to me, and possibly invalidates my findings on whether we need priv-scc for the operator to deploy openshift-pipelines

Would it be possible for one of you all to get specific/precise on those Makefile targets
```
opo-image - build image

opo-operator-yaml-update - updates the new image in the deployment manifest
```

instead of me having to hunt through all the *.mk files.  And the descriptions on the targets from `make help` are not proving helpful to this first time user :-)  Among other things none of the opo-* targets are even listed

Basically, what exact command do you run from your local branch to "run using the deployment manifest, RBAC and Service Account in the deploy folder"

And I'll then just swap out any registries, image names under your IDs with mine.

Thanks.

gabemontero:
OK @nikhil-thomas I got make opo-image and make opo-operator-yaml-update working, with my version of the image pushed to my personal quay.io repo, and the operator.yaml2

I'm still unclear, does running `operator-sdk run local --watch-namespace ""  --operator-flags '--zap-encoder=console --recursive'`
now occur *WITHOUT* kubeadmin privs ?

Or do I need to use another form of `operator-sdk run` when doing this?

Although when I run that operator-sdk run local command either with or without the make changes, I'm now getting start up errors I as not getting before, and don't understand why

```
2020-06-29T14:08:39.183-0400	INFO	ctrl.config.create-cr	creating a clusterwide resource of config crd	{"name": "cluster"}
2020-06-29T14:08:41.921-0400	ERROR	ctrl.config.add	creation of config resource failed	{"error": "no matches for kind \"Config\" in version \"operator.tekton.dev/v1alpha1\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/tektoncd/operator/pkg/controller/config.add
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/pkg/controller/config/config_controller.go:171
github.com/tektoncd/operator/pkg/controller/config.Add
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/pkg/controller/config/config_controller.go:52
github.com/tektoncd/operator/pkg/controller.AddToManager
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/pkg/controller/controller.go:13
main.main
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/cmd/manager/main.go:115
runtime.main
	/usr/local/go/src/runtime/proc.go:203
2020-06-29T14:08:41.921-0400	ERROR	cmd		{"error": "no matches for kind \"Config\" in version \"operator.tekton.dev/v1alpha1\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/vendor/github.com/go-logr/zapr/zapr.go:128
main.main
	/home/gmontero/go/src/github.com/openshift/tektoncd-pipeline-operator/cmd/manager/main.go:116
runtime.main
	/usr/local/go/src/runtime/proc.go:203
FATA[0006] Failed to run operator: failed to run operator locally: failed to exec []string{"build/_output/bin/tektoncd-pipeline-operator-local", "--zap-encoder=console", "--recursive"}: exit status 1 

```

Would something have changed in the repo that I need to adjust to?  Do I need to update my operator-sdk binary ??

I'm stuck / need help.

sthaha:
> I'm stuck / need help.

I am pretty sure this occurs if you have forgotten to apply/add the CRD to the cluster.
Did you forget to `oc apply -f deploy/crds/operator_v1alpha1_config_crd.yaml   ` ?

sthaha:
Just incase it helps I am using ... 
```
Go Version: go1.13.3
Go OS/Arch: linux/amd64
Version of operator-sdk: v0.14.1
```
☝ is from the logs

gabemontero:
ok @nikhil-thomas @sthaha I was able to use this non operator-sdk approach to launch the operator:

```
oc apply -f deploy/crds/operator_v1alpha1_config_crd.yaml
export VERSION=dev
export QUAY_NAMESPACE=gabemontero
make opo-image
docker push quay.io/gabemontero/openshift-pipelines-operator:vdev
make opo-operator-yaml-update  # deploy/operator.yaml is updated
oc apply -f ./deploy/role.yaml
oc apply -f ./deploy/role_binding.yaml
oc apply -f ./deploy/service_account.yaml
oc apply -f ./deploy/operator.yaml 
```

per a video conf with @nikhil-thomas today, that should ensure the operator is not running as kubeadmin

I was able to
- verify openshift-pipelines is created, and the controllers are up
- create a test project and see the pipeline SA created
- create some simple clustertask/taskrun objs and see them execute correctly
- ran `oc adm policy who-can use scc privileged` and see that the SA `tekton-pipelines-controller` in the `openshift-pipelines` namespace is not listed as hoped for
- now, the `openshift-pipelines-operator` in the `openshift-operators` namespace is listed, based on the broad spectrum of permissions is https://github.com/openshift/tektoncd-pipeline-operator/blob/master/deploy/role.yaml
- that is not as big of a concern since it is not what converts users' pipelines/tasks into pods
- but I'll experiment with fine tuning that role definition, further restrict what it can do, but still do the above, and report back here with updates to the PR as needed

So the net:
- the earlier findings in that we only need the privilege scc for image building still hold, and we can do that via augmenting the SAs in question vs. the controller, as well as track the work we have on the build api team side / buildah team to reduced what is needed there

@vdemeester @chmouel @siamaksade @adambkaplan @sbose78 FYI

gabemontero:
OK I was able to narrow the permissions somewhat for the operator ... pushed as a second commit @nikhil-thomas @sthaha 

It might be time for ourselves and @piyush-garg to have a voice to voice conversation on where to go next ... a few possibilites
- add elevated priv to the pipeline SA to facilitate building images 
- where we either go with https://github.com/openshift/tektoncd-pipeline-operator/blob/master/deploy/resources/devel/pipelines/01-clusterrole.yaml or we explore some of the build api team's findings around the minimal escalations needed to build images with buidlah ... see https://gist.github.com/otaviof/e839e744e868653ac0439b8419ee4343
- determine the set of samples and doc to modify to conform to best practices around given specific SAs priv ... is the complete set:
-- https://github.com/openshift/tektoncd-pipeline-operator/tree/master/deploy/resources/devel/optional 
-- https://github.com/openshift/tektoncd-catalog/tree/release-v0.11
-- or are there others ? I assumed tektoncd-catalog was what we did instead of the upstream catalog ...is that right?

gabemontero:
Official implementation done with https://github.com/openshift/tektoncd-pipeline-operator/pull/503 by @sbose78

PR #340, opened 2020-06-26
Refactoring to use manifestival/manifestival

piyush-garg:
This will bump from jcrossley3/manifestival to
manifestival/manifestival as that repo is
archived now


pkg/controller/config/config_controller.go
```diff
@@ -99,38 +106,31 @@ func fetchCommuntiyResources(mgr manager.Manager) (mf.Manifest, error) {
...
-	if err != nil {
-		return mf.Manifest{}, err
-	}
-	addons.Resources = append(addons.Resources, optionalResources...)
-
 	return addons, nil
 }
 
-func readOptional(mgr manager.Manager) ([]unstructured.Unstructured, error) {
+func readOptional(mgr manager.Manager) (mf.Manifest, error) {
```
sthaha, line 116: would `*mf.Manifest` be better so that you can `return nil, err` ?

pkg/controller/config/config_controller.go
```diff
@@ -354,11 +372,11 @@ func (r *ReconcileConfig) applyCommunityResources(req reconcile.Request, cfg *op
 	}
 	log.Info("successfully applied all non Red Hat resources")
 
-	err := r.updateStatus(cfg, op.ConfigCondition{Code: op.InstalledStatus, Version: flag.TektonVersion})
+	err = r.updateStatus(cfg, op.ConfigCondition{Code: op.InstalledStatus, Version: flag.TektonVersion})
 	return reconcile.Result{Requeue: true}, err
 }
 
-func transformManifest(cfg *op.Config, m *mf.Manifest, addnTfrms ...mf.Transformer) error {
+func transformManifest(cfg *op.Config, m *mf.Manifest, addnTfrms ...mf.Transformer) (mf.Manifest, error) {
```
sthaha, line 379: how about we return `*mf.Manifest, error`?

sthaha: ok .. scratch that .. `m.Transform` returns mf.Manifest

piyush-garg: yes that is the change happened in dependency

PR #353, opened 2020-07-09
Change Buildah and S2I Image to Productized

khrm:
Updated update-tasks.sh script to change buildah and s2i image to productized version. Script can be used to change images of
other tasks to productized version also if needed.

### DIFF
--- a/scripts/update-tasks.sh
+++ b/scripts/update-tasks.sh
@@ -87,6 +87,22 @@ EOF
 
 }
 
+change_task_image() {
+  local dest_dir="$1"; shift
+  local version="${1//./-}"; shift
+
+  local task="$1"; shift
+  local task_path="$dest_dir/${task}/${task}-task.yaml"
+
+  local expr=$1; shift
+  local image=$1; shift
+  local loc=$1; shift
+
+  sed \
+      -i "s'$expr.*'$loc $image'" \
+      $task_path
+}
+
 get_tasks() {
   local dest_dir="$1"; shift
   local version="${1//./-}"; shift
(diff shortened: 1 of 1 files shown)


piyush-garg:
I think we do the image replacement in code, also kind change. We can do this also there

savitaashture:
Yes i agree with @piyush-garg that we should be able to handle image replacement in code like the way we handled for kn https://github.com/openshift/tektoncd-pipeline-operator/pull/308

khrm:
@piyush-garg @savitaashture 
We do `kind` change in script for Red Hat supported clustertasks.  Any tasks downloaded in the addons have their kind already fixed.

Reference: 
[update-tasks.sh#L81](https://github.com/openshift/tektoncd-pipeline-operator/blob/master/scripts/update-tasks.sh#L81)
[buildah-task.yaml#L8](https://github.com/openshift/tektoncd-pipeline-operator/blob/master/deploy/resources/devel/addons/03-clustertasks/buildah/buildah-task.yaml#L8)

nikhil-thomas:
> Yes i agree with @piyush-garg that we should be able to handle image replacement in code like the way we handled for kn #308

we can do this in 2 ways. 

1. #308 : images are replaced right before the operator creates the resource on cluster.
2. as in this patch, in replace images in the static files (resources)

what we do in #308 will still work on buildah and s2i even after we merge this patch.
In addition, this patch will make sure that the productized image is there in this codebase, so anyone using/building the operator can use it without having to specify an override in operator.yaml (or in csv) if they prefer that (default will be official images).

So
/approve
/lgtm

PR #365, opened 2020-07-23
CommunityResourceURLs: update according to catalog organization…

vdemeester:
- Track master branch
- Use versionned tasks

/cc @piyush-garg @nikhil-thomas @sthaha 

See https://issues.redhat.com/browse/SRVKP-825

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/pkg/flag/flag.go
+++ b/pkg/flag/flag.go
@@ -51,11 +51,11 @@ var (
 	Recursive              bool
 	OperatorUUID           string
 	CommunityResourceURLs  = []string{
-		"https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/jib-maven/jib-maven.yaml",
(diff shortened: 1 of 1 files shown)


chmouel:
/lgtm

Looking Good! 🤙🏽

piyush-garg:
i think we discussed to point to a specific commit sha

vdemeester:
> i think we discussed to point to a specific commit sha

We should have the guarantee, with the `version` to not require to target a commit sha. We were also thinking of using a fork, but I wonder if we should do that now or as a follow-up (or not).

/cc @sthaha

nikhil-thomas:
/lgtm
/approve

i think the version dir should be enough guarantee.
in addtion, not peggng to a commit will help us get updates/fixes in the tasks with the same version.

piyush-garg:
I see, but for the release branch, we should point to specific sha. This is for master, so I think fine to be the latest.

/hold cancel

PR #376, opened 2020-07-29
Handle Deployment Label Changes for Triggers and Pipelines

khrm:
Added Capability to handle labels in deployment of webhooks, controller deployments.
    We first delete deployment and then apply it.


vdemeester:
@khrm needs a rebase :sweat:

PR #381, opened 2020-07-31
Update clustertaks for v0.14.3

piyush-garg:
This will update all the tasks to fetch from catalog
v0.14.3 and refactor the script a bit to handle the
org changes.
Fetch both workspace and pr based task from pipelines
catalog
Move git-clone tasks from community to supported one.
Update the script to update the image in both task
and versioned task
Update the script to replace image in pipelines
catalog task also


scripts/update-tasks.sh
```diff
@@ -189,6 +202,18 @@ main() {
     "s2i"  "image: quay.io/buildah"  \
```
khrm, line 202: Can we remove this? Does it work then? It can come from the array, OPENSHIFT_CATALOG_TASKS also. Except buildah task which has default, other can from that array.

piyush-garg: s2i is in `TEKTON_CATALOG_TASKS`, not `OPENSHIFT_CATALOG_TASKS`

khrm: OK.

PR #386, opened 2020-08-04
Update pipelines

piyush-garg:
This will update pipeline to default to
be workspace based and use -pr for
pipeline resource based

### DIFF
--- a/deploy/resources/devel/addons/04-pipelines/deployment-config/buildah/buildah-build-deploy.yaml
+++ b/deploy/resources/devel/addons/04-pipelines/deployment-config/buildah/buildah-build-deploy.yaml
@@ -13,25 +13,44 @@ spec:
   params:
     - name: APP_NAME
       type: string
-  resources:
-    - name: app-source
-      type: git
-    - name: app-image
-      type: image
-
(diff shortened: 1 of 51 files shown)


deploy/resources/devel/addons/04-pipelines/deployment-config/buildah/buildah-build-deploy.yaml
```diff
@@ -13,25 +13,40 @@ spec:
   params:
     - name: APP_NAME
       type: string
-  resources:
-    - name: app-source
-      type: git
-    - name: app-image
-      type: image
-
+    - name: GIT_REPO
```
vdemeester, line 16: Do we want to allow passing the revision of the git repo in addition to the url ?

vdemeester: (as it is possible through PR)

piyush-garg: PR updated

PR #388, opened 2020-08-04
Update optional resource

piyush-garg:
This will update optional resources to be
have snippets of both workspace and pipeline
resource

### DIFF
--- a/deploy/resources/devel/optional/pipelinesamples/docker_build_deploy.yaml
+++ b/deploy/resources/devel/optional/pipelinesamples/docker_build_deploy.yaml
@@ -14,30 +14,47 @@ spec:
     apiVersion: tekton.dev/v1beta1
     kind: Pipeline
     metadata:
-      name: docker-build-and-deploy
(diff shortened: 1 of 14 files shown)


deploy/resources/devel/optional/pipelinesamples/docker_build_deploy.yaml
```diff
@@ -14,30 +14,43 @@ spec:
...
       params:
         - name: IMAGE_NAME
           type: string
-      resources:
-        - name: app-source
-          type: git
-        - name: app-image
-          type: image
-
+        - name: GIT_REPO
```
vdemeester, line 22: Same question as in https://github.com/openshift/tektoncd-pipeline-operator/pull/386#pullrequestreview-460780261. Shouldn't we also allow the user to pass a revision ?

piyush-garg: PR updated according to comments

PR #389, opened 2020-08-04
Consistent status

piyush-garg:
This will make the status consistent for all
the update status done in controller


piyush-garg:
This is basically fixing the inconsistency appeared after https://github.com/openshift/tektoncd-pipeline-operator/pull/376

PR #393, opened 2020-08-07
Add Preserve Rolebinding Subject Namespace

khrm:


deploy/resources/devel/pipelines/00-release.yaml
```diff
@@ -1722,6 +1722,8 @@ kind: RoleBinding
 metadata:
   name: openshift-pipelines-prometheus-k8s-read-binding
   namespace: tekton-pipelines
+  annotations:
+    operator.tekton.dev/preserve-rb-subject-namespace: "true"
```
nikhil-thomas, line 1726: let us avoid editing the pipeline release.yaml.

Instead of filtering and conditionally changing the subject namespace, could we use better filtering.
may be using manifestival provides ByName. In addition, can we write a custom `predicate` if required.

pkg/controller/config/config_controller.go
```diff
@@ -497,15 +498,32 @@ func (r *ReconcileConfig) applyCommunityResources(req reconcile.Request, cfg *op
...
+	rest, err := rest.Transform(tfs...)
+	if err != nil {
+		return *m, err
+	}
+
+	tfs = []mf.Transformer{
+		mf.InjectOwner(cfg),
+		transform.InjectNamespaceRoleBindingConditional(flag.AnnotationPreserveNS,
+			flag.AnnotationPreserveRBSubjectNS, cfg.Spec.TargetNamespace),
+		transform.InjectDefaultSA(flag.DefaultSA),
```
nikhil-thomas, line 520: do we need `InjectDefaultSA` while transforming RoleBindings

pkg/utils/transform/transform.go
```diff
@@ -69,12 +69,30 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
...
+	}
+}
+
 func InjectNamespaceRoleBindingSubjects(targetNamespace string) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		kind := strings.ToLower(u.GetKind())
 		if kind != "rolebinding" {
 			return nil
 		}
+		u.SetNamespace(targetNamespace)
```
nikhil-thomas, line 95: this is a shortcut, and could be confusing. here we are making a function with some other resposibility one additional thing. let us try to handle this differently.

In addition, why don't we keep `transform.InjectNamespaceRoleBindingSubjects(cfg.Spec.TargetNamespace),` in rbTransforms as well.

pkg/utils/transform/transform.go
```diff
@@ -69,12 +69,30 @@ func InjectNamespaceConditional(preserveNamespace, targetNamespace string) mf.Tr
 	}
 }
 
+func InjectNamespaceRoleBindingConditional(preserveNS, preserveRBSubjectNS, targetNamespace string) mf.Transformer {
```
nikhil-thomas, line 72: let us make this function check only `preserveRBSubjectNS` check
and reuse `InjectNamespaceRoleBindingSubjects` for `preserveNamespace Check

pkg/utils/transform/transform.go
khrm: That's there by mistake. Deleting it.

deploy/resources/devel/pipelines/00-release.yaml
khrm: Actually, this pipeline.yaml would come from https://github.com/openshift/tektoncd-pipeline/pull/485 here. So we aren't editing it but synchronising it with what we should have.

nikhil-thomas: oh, i see..  cool :+1:

pkg/utils/transform/transform.go
khrm: I didn't understand this. Why should we check only one thing?

khrm: I didn't get this. So we should have only one check there?

pkg/controller/config/config_controller.go
khrm: No.

pkg/controller/config/config_controller.go
```diff
@@ -497,15 +498,31 @@ func (r *ReconcileConfig) applyCommunityResources(req reconcile.Request, cfg *op
...
-	return m.Transform(tfs...)
+	rest, err := rest.Transform(tfs...)
+	if err != nil {
+		return *m, err
+	}
+
+	tfs = []mf.Transformer{
+		mf.InjectOwner(cfg),
+		transform.InjectNamespaceRoleBindingConditional(flag.AnnotationPreserveNS,
+			flag.AnnotationPreserveRBSubjectNS, cfg.Spec.TargetNamespace),
```
nikhil-thomas, line 519: can we add 
`transform.InjectNamespaceConditional(flag.AnnotationPreserveNS, cfg.Spec.TargetNamespace),` 
and `transform.InjectNamespaceRoleBindingConditional(flag.AnnotationPreserveRBSubjectNS, cfg.Spec.TargetNamespace),` 

and handle metadata Namespace and subject->Namespace separately in 2 different transformers

pkg/utils/transform/transform.go
nikhil-thomas: https://github.com/openshift/tektoncd-pipeline-operator/pull/393/files?file-filters%5B%5D=.go#r467888300

pkg/controller/config/config_controller.go
khrm: Problem is that transform.InjectNamespaceConditional would transform the namespace of subject also because it calls manifestival transform.InjectNamespace which transforms subject namespace.

khrm: https://github.com/manifestival/manifestival/blob/17489fb533562ae0e7c47c14a4ac9bfe55e8af8c/transform.go#L66

It update subject here

pkg/utils/transform/transform.go
nikhil-thomas: it makes sense now

nikhil-thomas:
cc @vdemeester

PR #404, opened 2020-08-13
Disable affinity assistant feature

piyush-garg:
This will disable the affinity assistant feature in
pipelines by putting the value of
disable-affinity-assistant flag in feature-flags
configmap to true


piyush-garg:
/approve

Just triggering tide

PR #408, opened 2020-08-18
Fix permissions for operator

piyush-garg:
This will fix the issue of operator not able to
install pipeline because of changes in
monitoring yaml and during upgrade it
looks like update not create


sthaha (commented):
/lgtm
/approve
/meow

PR #410, opened 2020-08-19
Fix for smooth operator upgrades

piyush-garg:
This will add the functionality of recreating
services also alog with deployment

Separate out triggers from addons
As we need to make triggers up
before applying addons, also
add check for validating triggers

Related test, script and resource changes


khrm (approved):
/lgtm

PR #412, opened 2020-08-20
Add functional to delete triggers

piyush-garg:
As we moved triggers into separate resource from addons
the deleteFunction missed for that change and
triggers are not getting deleted on uninstall


vdemeester (approved):
/lgtm

PR #413, opened 2020-08-20
runtime version support for +Add flow in console

pradeepitm12:
This adds runtime version support for +Add flow in console by
adding all possible pipeline with different versioned clustertaks

Changes
 - Create all the pipeline with different builder and different
   environment
 - Pipeline creation done with help of online one pipeline teplate
 - No need to maintain same copy of pipelines with minor changes
 - adds "pipeline.openshift.io/supported-versions" annotation
   which helps console to work with correct runtime version


pkg/controller/config/config_controller.go
```diff
@@ -722,3 +731,20 @@ func sourceBasedOnRecursion(path string) mf.Source {
 	}
 	return mf.Path(path)
 }
+
+func createAddonsPipelines(mgr manager.Manager) (mf.Manifest, error) {
+	addonsPath := filepath.Join("pkg/utils/addons", "pipelinetemplate.yaml")
```
nikhil-thomas, line 736: shal we move this template also in to `deploy/resources/addons` as this is something which might change in each release.

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,215 @@
...
+		"s2i-dotnet-3": {runtime: "dotnet", minorVersion: "1", supportedVersion: "[3.1,3.0]"},
+		"s2i-go":       {runtime: "golang"},
+		"s2i-java-8":   {runtime: "java", supportedVersion: "[8]"},
+		"s2i-java-11":  {runtime: "java", supportedVersion: "[11]"},
+		"s2i-nodejs":   {runtime: "nodejs", version: "10", supportedVersion: "[10,12]"},
+		"s2i-perl":     {runtime: "perl", minorVersion: "26", supportedVersion: "[5.26,5.24]"},
+		"s2i-php":      {runtime: "php", minorVersion: "2", supportedVersion: "[7.2,7.3]"},
+		"s2i-python-3": {runtime: "python", minorVersion: "6", supportedVersion: "[3.6,3.5]"},
+		"s2i-ruby":     {runtime: "ruby", minorVersion: "5", supportedVersion: "[2.5,2.4,2.3]"},
+	}
```
nikhil-thomas, line 34: let us move this config into `pkg/flag.go`

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,215 @@
...
+	version          string
+	minorVersion     string
+	supportedVersion string
+}
+
+var (
+	runtimeLabel                = "pipeline.openshift.io/runtime"
+	typeLabel                   = "pipeline.openshift.io/type"
+	supportedVersionsAnnotation = "pipeline.openshift.io/supported-versions"
+	preserveNSAnnotation        = "operator.tekton.dev/preserve-namespace"
```
nikhil-thomas, line 19: some of these are already there in `pkg/flag.go`. 
Let us reuse or move the additional ones into `pkg/flag/flag.go`

pkg/controller/config/config_controller.go
```diff
@@ -722,3 +731,20 @@ func sourceBasedOnRecursion(path string) mf.Source {
...
+	if err != nil {
+		return mf.Manifest{}, err
+	}
+	pipelines, err := addons.CreatePipelines(pipelineTemplate)
+	if err != nil {
+		return mf.Manifest{}, err
+	}
+	m1, err := mf.ManifestFrom(mf.Slice(pipelines),mf.UseClient(mfc.NewClient(mgr.GetClient())))
+	if err != nil {
+		return mf.Manifest{}, err
```
nikhil-thomas, line 747: could you move this like to the above function, and make `addons.CreatePipeliens` return `manifest`

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,215 @@
...
+		taskDeployMap["taskRef"] = map[string]interface{}{"name": "kn", "kind": "ClusterTask"}
+		taskDeployMap["runAfter"] = []interface{}{"build"}
+
+		taskDeployMap["params"] = []interface{}{
+			map[string]interface{}{"name": "ARGS", "value": []interface{}{"service", "create", "$(params.APP_NAME)", "--image=$(params.IMAGE_NAME)", "--force"}},
+		}
+		pipelines = append(pipelines, newTempRes)
+	}
+	return pipelines, nil
+}
```
nikhil-thomas, line 215: majority of the logic in these functions are the same. 
could you refactor these three functions, we can reuse common functionalities
func-diff

pkg/controller/config/config_controller.go
pradeepitm12: I think putting `pipelinetemplate.yaml` in   `deploy/resources/addons` will cause issues because this template is an incomplete yaml and when operator try to apply this yaml will give an error

pkg/utils/addons/addons_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+		kTemplate, err := createKnativePipeline(template, runtime)
+		assertNoEror(t, err)
+
+		jsonkTemp, err := kTemplate[0].MarshalJSON()
+		assertNoEror(t, err)
+
+		actualdata += string(jsonkTemp) + "\n"
+
+		golden.Assert(t, actualdata, strings.ReplaceAll(fmt.Sprintf("%s.golden", t.Name()), "/", "-"))
```
nikhil-thomas, line 44: very nice :+1: :heart:

pkg/utils/addons/addons.go
pradeepitm12: Thank you, will update it.

pkg/utils/addons/addons_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	mf "github.com/manifestival/manifestival"
+	"gotest.tools/golden"
+)
+
+func TestCreatePipeline(t *testing.T) {
+	t.Run("dotnet pipeline", func(t *testing.T) {
+		var runtime = map[string]spec{
+			"s2i-dotnet-3": {runtime: "dotnet", minorVersion: "1", supportedVersion: "[3.1,3.0]"},
+		}
```
nikhil-thomas, line 16: can we make the three tests in this file into a table based test

```
runtime[] = {
map[string]spec{
			"s2i-dotnet-3": {runtime: "dotnet", minorVersion: "1", supportedVersion: "[3.1,3.0]"},
		},
map[string]spec{
			"s2i-nodejs": {runtime: "nodejs", version: "10", supportedVersion: "[10,12]"},
		}
...
}
```

then 
```
for runtime in runtimes
    run tests
```

nikhil-thomas (changes requested):

nikhil-thomas:
Nice work :+1: 
I love the fact that the number of static files got reduced from 27 to 1. It will be easier to maintain.
I have made few suggestions. Could you address those.

pkg/utils/addons/addons_test.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	mf "github.com/manifestival/manifestival"
+	"gotest.tools/golden"
+)
+
+func TestCreatePipeline(t *testing.T) {
```
vdemeester, line 12: Could we use test data table for those ? (aka a loop with `t.Run`)

pkg/controller/config/config_controller.go
nikhil-thomas: i see, 
then could you place it in `deploy/resources/04-templates`

pradeepitm12: Yes, this can be done.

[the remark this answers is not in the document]
pradeepitm12: I too had the same thought, followed it from earlier code. But now we are using the path straight forward.

pkg/utils/addons/addons_test.go
pradeepitm12: Thanks for pointing this, earlier we faced an issue while testing as the order was not maintained. But the new code takes care of this.

pkg/utils/addons/addons_test.go
pradeepitm12: With new implementations, this is completely changed. Thank you for the suggestion.

pkg/controller/config/config_controller.go
```diff
@@ -828,3 +837,16 @@ func sourceBasedOnRecursion(path string) mf.Source {
 	}
 	return mf.Path(path)
 }
+
+func createAddonsPipelines(mgr manager.Manager) (mf.Manifest, error) {
+	pipelineTemplate, err := mf.NewManifest("deploy/resources/templates/pipelinetemplate.yaml")
```
nikhil-thomas, line 842: move this path to `pkg/flag.go` and make you of `path.Join()`

pkg/flag/flag.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/spf13/pflag"
 )
 
+type RuntimeSpec struct {
+	Runtime           string
+	Version           string
+	MinorVersion      string
+	SupportedVersions string
+}
+
```
nikhil-thomas, line 17: :+1:

pkg/flag/flag.go
```diff
@@ -39,6 +46,10 @@ const (
 	ProviderTypeRedHat            = "redhat"
 	ProviderTypeCertified         = "certified"
 
+	AnnotationSupportedVersions = "pipeline.openshift.io/supported-versions"
+	LabelEnvironmentType        = "pipeline.openshift.io/type"
+	LabelPipelineRuntime        = "pipeline.openshift.io/runtime"
+
```
nikhil-thomas, line 52: for better context, let us add the word `Pipeline` in Supported versions and enviorment as well
eg: `LabelPipelineEnviormentType`

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,166 @@
...
+
+type deploymentPipeline struct {
+	environment string
+	nameSuffix  string
+}
+
+type knativePipeline struct {
+	environment string
+	nameSuffix  string
+}
```
nikhil-thomas, line 28: could they be objects of a single struct rather than separate structs.
can we just move the part which defer in the implementation into interface methods

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (dcp *dcPipeline) generate(pipeline unstructured.Unstructured) (unstructured.Unstructured, error) {
+	newTempRes := unstructured.Unstructured{}
+	pipeline.DeepCopyInto(&newTempRes)
+	labels := newTempRes.GetLabels()
+	labels[flag.LabelEnvironmentType] = dcp.environment
+	newTempRes.SetLabels(labels)
```
nikhil-thomas, line 35: this par is common in all generate functions..

we just need to move `the parts which` are different into individual methods

pkg/utils/addons/addons.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func CreatePipelines(template mf.Manifest, client client.Client) (mf.Manifest, error) {
+	var pipelines []unstructured.Unstructured
+
+	pipelineGenerators := []pipelineGenerator{
+		&dcPipeline{environment: "openshift", nameSuffix: ""},
+		&deploymentPipeline{environment: "kubernetes", nameSuffix: "-deployment"},
+		&knativePipeline{environment: "knative", nameSuffix: "-knative"},
+	}
```
nikhil-thomas, line 104: `{environment: "openshift", nameSuffix: ""}` 
it is the same struct.

what we need here is a single struct,
and 
which takes in 
- enviorment
- nameSuffix
- and an interface type object which defines the difference in the pipelines

nikhil-thomas:
@pradeepitm12 Thanks for the updates :+1: 

we could make the code in ` pkg/utils/addons/addons.go` bit more simpler.
Could you take one more look at it. 
I have added few comments on the file.

pkg/controller/config/config_controller.go
pradeepitm12: I used the path.Join() earlier but then directly used complete path, do we have any specific reason to use path.Join().

pkg/flag/flag.go
pradeepitm12: This actually makes sense, will update this. Thank you

pkg/utils/addons/addons.go
pradeepitm12: So all the generating functions are adding third task which is **deploy** in its own way, we do see some of the similar parts in that task. And I wanted to keep this task separately. I will check if I can move things to a different function, something like which generate the base of the task.

pkg/controller/config/config_controller.go
nikhil-thomas: yes, people might run tests from different platforms, where platforms are `/` or `\`

pkg/utils/addons/addons.go
nikhil-thomas: the three struct has the same fields
```
environment string
nameSuffix  string
```
then why do we need three structs for same type of objects...


so what we need, is one type (`struct`) with the same fields 
```
environment string
	nameSuffix  string

Comment (nikhil-thomas) [reply pkg/utils/addons/addons.go:35]: atleast, let us create a type `pipeline`
```
type pipeline struct {
       environment string
	nameSuffix  string
}
```

then embedd it

eg: 
```
type deploymentPipeline struct {
	pipeline
}
```

Comment (nikhil-thomas) [reply pkg/utils/addons/addons.go:35]: or what we could do is 

create a type
```
type Pipeline struct {
       environment string
	nameSuffix  string
       pipelineGenerator
}
```

```
eg: 
type dcGenerator struct {}

func (d dcGenerator) generate() {
}
```

the `pipelineGenerator` type could be `interface type` or a `function type`

and then in create pipelines

```pipelines := []pipelines{
		&Pipeline{environment: "openshift", nameSuffix: "", dcGenerator{}},
		&Pipeline{environment: "kubernetes", nameSuffix: "-deployment", deploymentGenerator},
		&Pipeline{environment: "knative", nameSuffix: "-knative", knativeGenerator},
	}
```

PR #414, opened 2020-08-20
Delete and create addons

piyush-garg:
This will add the functionality of delete and create
addons because oc apply is not working
perfectly if you want apply same resource
again with some changes


vdemeester (approved):
/meow
/lgtm

pkg/controller/config/config_controller.go
```diff
@@ -284,41 +288,33 @@ func (r *ReconcileConfig) applyPipeline(req reconcile.Request, cfg *op.Config) (
...
+			Version: flag.TektonVersion})
 		return reconcile.Result{}, fmt.Errorf("failed to apply non deployment and service pipeline manifest: %w", err)
 	}
 	if err := r.pipeline.Filter(recreateResource).Apply(); err != nil {
 		if errors.IsInvalid(err) {
 			if err := r.deleteAndCreate(); err != nil {
 				_ = r.updateStatus(cfg, op.ConfigCondition{
-					Code:            op.PipelineApplyError,
-					Details:         err.Error(),
-					PipelineVersion: pipelineVersion,
```
khrm, line 300: Why did we remove this?

piyush-garg: it is not required field anymore, so we did wherever the field should be put like pipleineVersion after validate-pipeline and triggerVersion after validate-trigger. Also we were reading the label at the particular stage, now we moved it to start as it resolves the case of pod to restarted from the state after validate-triggers it doe snot show version then

PR #438, opened 2020-09-09
bump operator-sdk version to v0.17.2

savitaashture:
bump operator-sdk version to v0.17.2
Fixed Lint issues

Signed-off-by: Savita Ashture <sashture@redhat.com>


make/test.mk
```diff
@@ -139,7 +139,7 @@ test-e2e-ci: get-test-namespace ./vendor
 	$(Q)sed -e 's|REPLACE_NAMESPACE|$(TEST_NAMESPACE)|g' ./deploy/test/role_binding_test.yaml | oc apply -f -
 	$(Q)sed -e 's|REPLACE_IMAGE|registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:tektoncd-pipeline-operator|g' ./deploy/test/operator_test.yaml  | oc apply -f - --namespace $(TEST_NAMESPACE)
 	$(eval DEPLOYED_NAMESPACE := $(TEST_NAMESPACE))
-	$(Q)operator-sdk test local ./test/e2e --namespace $(TEST_NAMESPACE) --no-setup --go-test-flags "-v -timeout=15m"
+	$(Q)operator-sdk test local ./test/e2e --operator-namespace $(TEST_NAMESPACE) --no-setup --go-test-flags "-v -timeout=15m"
```
nikhil-thomas, line 142: could you add `--watch-namespace ""` as well
we need both.

from v17.2 we have 2 flags instead of the old `--namespace` as you noticed.

if we don't set `--watch-namespace=""` the operator under test will only watch one namesapce, that will be = value of `--operator-namespace`

make/test.mk
```diff
@@ -192,7 +192,7 @@ else
 	$(Q)sed -i 's|$(TEST_NAMESPACE)|REPLACE_NAMESPACE|g' ./deploy/test/role_binding_test.yaml
 	$(Q)sed -i 's|172.30.1.1:5000/$(TEST_NAMESPACE)/tektoncd-pipeline-operator:latest|REPLACE_IMAGE|g' ./deploy/test/operator_test.yaml
 endif
-	$(Q)eval $$(minishift docker-env) && operator-sdk test local ./test/e2e --namespace $(TEST_NAMESPACE) --no-setup --go-test-flags "-v"
+	$(Q)eval $$(minishift docker-env) && operator-sdk test local ./test/e2e --operator-namespace $(TEST_NAMESPACE) --no-setup --go-test-flags "-v"
```
nikhil-thomas, line 195: same as above

opo-makefile.mk
```diff
@@ -65,11 +65,11 @@ endif
 
 .PHONY: opo-test-e2e-up-local
 opo-test-e2e-up-local: opo-test-clean
-	operator-sdk test local ./test/e2e/ --up-local --namespace openshift-pipelines  --go-test-flags "-v -timeout=10m" --local-operator-flags "--recursive"
+	operator-sdk test local ./test/e2e/ --up-local --operator-namespace openshift-pipelines  --go-test-flags "-v -timeout=10m" --local-operator-flags "--recursive"
```
nikhil-thomas, line 68: could you add `--watch-namespace ""` as well
we need both.

from v17.2 we have 2 flags instead of the old `--namespace` as you noticed.

if we don't set `--watch-namespace=""` the operator under test will only watch one namesapce, that will be = value of `--operator-namespace`

opo-makefile.mk
```diff
@@ -65,11 +65,11 @@ endif
...
 
 .PHONY: opo-test-e2e
 opo-test-e2e: opo-test-clean
-	operator-sdk test local ./test/e2e/ --namespace openshift-operators  --go-test-flags "-v -timeout=10m" --local-operator-flags "--recursive"
+	operator-sdk test local ./test/e2e/ --operator-namespace openshift-operators  --go-test-flags "-v -timeout=10m" --local-operator-flags "--recursive"
```
nikhil-thomas, line 72: here only `--operator-namesapce` is requried :+1: 
as there is no `--up-local` flag, the `WATCH_NAMESPCE` var is set from the deployment manifest `deploy/operator.yaml`

nikhil-thomas:
@savitaashture thanks for the quick fix :+1: 
the ci is failing because the ci still uses an old version operator-sdk cli to run tests.
could you update operator-sdk version in `openshift-ci/Dockerfile.tools` as well.
https://github.com/openshift/tektoncd-pipeline-operator/blob/3225a8366abda107633c9c3d06791de1e5d67307/openshift-ci/Dockerfile.tools#L65

nikhil-thomas:
cc @snehlatamohite

savitaashture:
> @savitaashture thanks for the quick fix 
> the ci is failing because the ci still uses an old version operator-sdk cli to run tests.
> could you update operator-sdk version in `openshift-ci/Dockerfile.tools` as well.
> https://github.com/openshift/tektoncd-pipeline-operator/blob/3225a8366abda107633c9c3d06791de1e5d67307/openshift-ci/Dockerfile.tools#L65

I have updated it https://github.com/openshift/tektoncd-pipeline-operator/pull/438/files?file-filters%5B%5D=#diff-48abca7cd218705cd020054b38fa29cdR65 :thinking:

make/test.mk
savitaashture: I think `--watch-namespace` is required if there is a `--up-local` flag as per this PR https://github.com/operator-framework/operator-sdk/pull/2617 information

opo-makefile.mk
savitaashture: Updated Thank you

nikhil-thomas:
@khrm @piyush-garg

PR #453, opened 2020-09-21
autogenerate pipelineresources based pipelines

pradeepitm12:
this adds autogeneration for pipelineresource based pipelines
and brings down another 30 static file to keep and maintain.


jerolimov:
@pradeepitm12 Just one recommendation to make future changes easier to track. Can you format the golden json files? I expect this makes diffs / PR reviews much easier in the future if there are any changes which effects this files, or?

pradeepitm12:
@jerolimov This actually makes sense to me, I will try to get that either formated or a YAML of it.

pkg/controller/config/config_controller.go
```diff
@@ -839,11 +839,7 @@ func sourceBasedOnRecursion(path string) mf.Source {
 }
 
 func createAddonsPipelines(mgr manager.Manager) (mf.Manifest, error) {
-	pipelineTemplate, err := mf.NewManifest("deploy/resources/templates/pipelinetemplate.yaml")
-	if err != nil {
-		return mf.Manifest{}, err
-	}
-	pipelines, err := addons.CreatePipelines(pipelineTemplate, mgr.GetClient())
+	pipelines, err := addons.CreatePipelines(flag.TemplatePath, mgr.GetClient())
```
nikhil-thomas, line 842: if we are passing `flag.TemplatePath` here, then shall we use `flag.TemplatePath` directly inside the function.
Is it necessary to to pass it from the controller.

pkg/utils/addons/addons.go
```diff
@@ -1,22 +1,26 @@
...
 
-type generateDeployTask func(interface{}) map[string]interface{}
-
 type pipeline struct {
-	environment string
-	nameSuffix  string
+	environment     string
+	nameSuffix      string
+	resourcedSuffix string
 	generateDeployTask
```
nikhil-thomas, line 24: could you add short comments for each of these fields. It could appear confusing to someone who wants to work on it in the future. I mean, it will be fantastic if someone can make sense of what each field is without having to read the rest of the code.

nikhil-thomas: `resourcedSuffix string` 
ok, after reading `createPipelines function, i think, we can handle this better.
what is the real purpose of this.

i see that there is a boolean parameter in
 `func generateBasePipeline(template mf.Manifest, taskGenerators []taskGenerator, resourced bool) ([]unstructured.Unstructured, error) {` 

and later
```
if resourced {
			taskName += "-pr"
		}
```

One way to handle this better is to add a function type, `type pipelineModifier func(map[string]interface{}) map[string]interface{}`

then, may be make

```
type pipeline struct {
	environment     string
	nameSuffix      string
	modifiers []pipelineModifier
```

Then generateDeployTask implemention(s) could be 1 item and for pipelineResource based ones there can be an additional modifiers

eg: 
```
workspacedTaskGenerators := []taskGenerator{
		&pipeline{environment: "openshift", nameSuffix: "", modifiers: []pipelineModifier{modiopenshiftDeployTask}},
		...
	}
```

and 
```
resourcedTaskGenerators := []taskGenerator{
		&pipeline{environment: "openshift", nameSuffix: "",modifiers: []pipelineModifier{modiopenshiftDeployTask, pipelineResourceModifications} openshiftDeployTask},
...	}
```

pkg/utils/addons/addons.go
```diff
@@ -121,26 +142,65 @@ func CreatePipelines(template mf.Manifest, client client.Client) (mf.Manifest, e
...
+
+	resourcedTemplate, err := mf.NewManifest(path.Join(templatePath, "pipeline_using_resource.yaml"))
+	if err != nil {
+		return mf.Manifest{}, err
+	}
+
+	resourcedTaskGenerators := []taskGenerator{
+		&pipeline{environment: "openshift", nameSuffix: "", resourcedSuffix: "-pr", generateDeployTask: openshiftDeployTask},
+		&pipeline{environment: "kubernetes", nameSuffix: "-deployment", resourcedSuffix: "-pr", generateDeployTask: kubernetesDeployTask},
+		&pipeline{environment: "knative", nameSuffix: "-knative", resourcedSuffix: "-pr", generateDeployTask: knativeResourcedDeployTask},
```
nikhil-thomas, line 192: why is only knative pipline needs a `knativeResourcedDeployTask` where as the other 2 seems to be reusing the `openshiftDepllyTask` and `kubernetesDeployTask`

pkg/utils/addons/addons.go
```diff
@@ -27,56 +31,65 @@ func (p *pipeline) generate(pipeline unstructured.Unstructured) (unstructured.Un
...
+	deployTask["resources"] = map[string]interface{}{
+		"inputs": []interface{}{map[string]interface{}{"name": "image", "resource": "app-image", "from": []interface{}{"build"}}},
+	}
+	deployTask["params"] = []interface{}{
+		map[string]interface{}{"name": "ARGS", "value": []interface{}{"service", "create", "$(params.APP_NAME)", "--image=$(resources.inputs.image.url)", "--force"}},
 	}
+	return deployTask
+}
+
+func generateBasePipeline(template mf.Manifest, taskGenerators []taskGenerator, resourced bool) ([]unstructured.Unstructured, error) {
```
nikhil-thomas, line 91: if we decide to go on with a boolean flag `resourced bool` , i would prefer a more descriptive name.
even `usePipelineResource bool` :)

pkg/utils/addons/addons.go
```diff
@@ -98,18 +111,26 @@ func CreatePipelines(template mf.Manifest, client client.Client) (mf.Manifest, e
...
+			return nil, err
+		}
+		var index int
+		if !resourced {
+			index = 1
+		}
+		taskBuild := tasks.([]interface{})[index].(map[string]interface{})
+		taskName := name
+		if resourced {
+			taskName += "-pr"
```
nikhil-thomas, line 128: if the default value of `index` is set to 1, then 
the 2 `if` blocks can be combined

pkg/controller/config/config_controller.go
nikhil-thomas: in fact the `createAddonsPipelines` is unnecessary in this update. 
In `readAddons` (in config_controller.go), we can directly call `addons.CreatePipelines`. Then read `falg.TemplatePath` from addons package.

pkg/utils/addons/addons.go
```diff
@@ -121,26 +142,65 @@ func CreatePipelines(template mf.Manifest, client client.Client) (mf.Manifest, e
...
 		}
 
 		if err := unstructured.SetNestedField(taskBuild, taskParams, "params"); err != nil {
-			return mf.Manifest{}, nil
+			return nil, nil
 		}
 
 		//adding the deploy task
 		for _, tg := range taskGenerators {
 			p, err := tg.generate(newTempRes)
```
nikhil-thomas, line 154: suggestion: if we consider all kinds of modifcations to pipelines as a modifier (`type pipelineModifier func(map[string]interface{}) map[string]interface{}`), then we could use a second loop to take care of various modifications here.

pkg/controller/config/config_controller.go
pradeepitm12: It was intentional to pass the template path, as I was facing issues while adding tests for it. I am not sure why the test was not able to read the resource in `deploy/resources/templates`.
Please let me know if there is something known.

pradeepitm12: Oh, you are right, we update the function. will remove it and call directly.

pkg/utils/addons/addons.go
pradeepitm12: right, I did this for the other one. Thanks

pkg/utils/addons/addons.go
pradeepitm12: deploy task for openshift, kubernetes environment does not have any change using resource or using the workspace, but for knative there is.

pkg/controller/config/config_controller.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/prometheus/common/log"
 	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 	"github.com/tektoncd/operator/pkg/flag"
-	"github.com/tektoncd/operator/pkg/utils/addons"
+	paddons "github.com/tektoncd/operator/pkg/utils/addons"
```
piyush-garg, line 18: we dont need an alias

pkg/utils/addons/addons.go
```diff
@@ -5,78 +5,89 @@ import (
 	mf "github.com/manifestival/manifestival"
 	"github.com/tektoncd/operator/pkg/flag"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"path"
```
piyush-garg, line 8: seems not go imported

pkg/utils/addons/addons_test.go
```diff
@@ -2,17 +2,15 @@ package addons
...
-
-	mf "github.com/manifestival/manifestival"
 	op "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 	"gotest.tools/golden"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"strings"
```
piyush-garg, line 12: not go imported

piyush-garg (changes requested):

PR #458, opened 2020-09-24
adds path_context param

pradeepitm12:
this adds path_context param in pipeline to be used in
build tasks

Note: for buildah task name of the param is CONTEXT
whereas for s2i tasks its PATH_CONTEXT to handle
this we have kept the pipeline param as PATH_CONTEXT and
change it at the task level like below

for s2i tasks
- name: PATH_CONTEXT
  value: $(params.PATH_CONTEXT)

for buildah
- name: CONTEXT
  value: $(params.PATH_CONTEXT)


vdemeester (commented):
/lgtm
/cc @savitaashture @khrm @nikhil-thomas

nikhil-thomas:
cc @jerolimov

jerolimov:
@pradeepitm12 @nikhil-thomas Great. :+1:

You're the experts here, did we need a default value or is the task default value (".") enough? I think the task default enough, or?

pradeepitm12:
@jerolimov Thanks for the review, I think we need to put the default value in here.
cc: @vdemeester

piyush-garg (approved):

PR #467, opened 2020-09-30
Refactor in clustertask provided type

piyush-garg:
This will mark the clustertask as provided by redhat for all
available in codebase and mark which are created during installation
as community

Also move the kn task from community to redhat, and the changes
regarding this in operator deployment and csv


khrm (approved):
/lgtm

PR #469, opened 2020-09-30
Add v1.2 Changes to flag pkg

khrm:


pkg/flag/flag.go
```diff
@@ -69,17 +69,17 @@ var (
...
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/tkn/0.1/tkn.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/kn/0.1/kn.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/helm-upgrade-from-source/0.1/helm-upgrade-from-source.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/helm-upgrade-from-repo/0.1/helm-upgrade-from-repo.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/skopeo-copy/0.1/skopeo-copy.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/trigger-jenkins-job/0.1/trigger-jenkins-job.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/git-cli/0.1/git-cli.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/pull-request/0.1/pull-request.yaml",
-		"https://raw.githubusercontent.com/tektoncd/catalog/master/task/kubeconfig-creator/0.1/kubeconfig-creator.yaml",
+		"https://raw.githubusercontent.com/tektoncd/catalog/755fed024504f8ed6ce89a338c3693e8a30de11b/task/jib-maven/0.1/jib-maven.yaml",
```
khrm, line 72: Why don't we use openshift/tektoncd-catalog here instead of tektoncd/catalog?

pkg/flag/flag.go
```diff
@@ -69,17 +69,17 @@ var (
...
+		"https://raw.githubusercontent.com/tektoncd/catalog/ec7c4f3836fe8ae4efd761e737e0b476a10c336d/task/maven/0.1/maven.yaml",
+		"https://raw.githubusercontent.com/tektoncd/catalog/a5e8f0cf6442378266fb353d6d4ad17f74903162/task/tkn/0.1/tkn.yaml",
+		"https://raw.githubusercontent.com/tektoncd/catalog/28a08e2e411c88ee20b2b600f5eb5ae893cf39c1/task/kn/0.1/kn.yaml",
+		"https://raw.githubusercontent.com/tektoncd/catalog/58ba68821eb85aa545195374a004311139eec097/task/helm-upgrade-from-source/0.1/helm-upgrade-from-source.yaml",
```
savitaashture, line 76: Why are we referring different commit id ?

pkg/flag/flag.go
piyush-garg: we mark them as community tasks

pkg/flag/flag.go
piyush-garg: We should refer to a single commit in master branch

pkg/flag/flag.go
nikhil-thomas: as we don't maintain them directly, the appropriate link would be the upstream/catalog link

pkg/flag/flag.go
nikhil-thomas: +1:

khrm: Done.

Issue #478, opened 2020-10-06
Skip installation of Upstream Community Tasks on Power/Z clusters


nikhil-thomas:
# Current Behavior

Upstream community tasks are installed on all OpenShift Clusters irrespective of the architecture.

Upstream Community Tasks are maintained by non Red Hat contributors. Hence, it is almost impossible to ensure that the step images in those tasks are supported on all architectures (x86,_64, ppc64le, s390x).

# Expected Behavior

Therefore the operator should check the architecture of the cluster (nodes) and skip installation of upstream community task if the platform is not x86_64

nikhil-thomas:
one way to handle this is to add a `platform` check in the `fetchCommunityResources`. We can skip the community resources if the platform is not `x86_64`.

https://github.com/openshift/tektoncd-pipeline-operator/blob/a59f72478cf7fc0aabad132fa9299b2dd7a99a48/pkg/controller/config/config_controller.go#L104

snehlatamohite:
@bahetiamit - Could you please have a look at it?

snehlatamohite:
@nikhil-thomas - Created PR https://github.com/openshift/tektoncd-pipeline-operator/pull/488 to skip upstream community tasks installation if arch is P OR Z.  Please review and do let me know if any changes are needed.

nikhil-thomas:
@snehlatamohite the `check` in you pr is correct, but it is in the `addons` block: https://github.com/openshift/tektoncd-pipeline-operator/blob/f2f3524f29632d9807fd19cc2feb864aede83ab6/pkg/controller/config/config_controller.go#L126

it should be in the `fetchCommunityResources` block: https://github.com/openshift/tektoncd-pipeline-operator/blob/f2f3524f29632d9807fd19cc2feb864aede83ab6/pkg/controller/config/config_controller.go#L105

could you make a patch.

@savitaashture @vdemeester

snehlatamohite:
@nikhil-thomas - Yes I will so the needful and will raise PR for required changes.

snehlatamohite:
@nikhil-thomas  - I have created PR https://github.com/openshift/tektoncd-pipeline-operator/pull/546 added check in fetchCommunityResources please let me know if I need to remove check from readAddons block?

PR #488, opened 2020-10-19
skip installation of tektoncd/catalog tasks if the platform is not x8…

snehlatamohite:
Hi All,

As requested in https://github.com/openshift/tektoncd-pipeline-operator/issues/478 ticket , I have created PR to skip installation of tektoncd/catalog tasks if arch is Ppc64le and s360x. 

Also updated Docker file since noticed now binary name is tektoncd-pipeline-operator . 

-COPY build/_output/bin/openshift-pipelines-operator ${OPERATOR}
+COPY build/_output/bin/tektoncd-pipeline-operator ${OPERATOR}

Please review and let me know if any changes are needed. 

Thanks,
Snehlata Mohite. 

Thanks,
Snehlata Mohite. 

Thanks,
Snehlata Mohite.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,7 +5,7 @@ ENV OPERATOR=/usr/local/bin/openshift-pipelines-operator \
     USER_NAME=openshift-pipelines-operator
 
 # install operator binary
-COPY build/_output/bin/tektoncd-pipeline-operator ${OPERATOR}
+COPY build/_output/bin/openshift-pipelines-operator ${OPERATOR}
 
 COPY build/bin /usr/local/bin
 RUN  /usr/local/bin/user_setup
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -5,7 +5,7 @@ ENV OPERATOR=/usr/local/bin/openshift-pipelines-operator \
     USER_NAME=openshift-pipelines-operator
 
 # install operator binary
-COPY build/_output/bin/openshift-pipelines-operator ${OPERATOR}
+COPY build/_output/bin/tektoncd-pipeline-operator ${OPERATOR}
 
 COPY build/bin /usr/local/bin
 RUN  /usr/local/bin/user_setup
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -5,7 +5,7 @@ ENV OPERATOR=/usr/local/bin/openshift-pipelines-operator \
     USER_NAME=openshift-pipelines-operator
 
 # install operator binary
-COPY build/_output/bin/openshift-pipelines-operator ${OPERATOR}
+COPY build/_output/bin/tektoncd-pipeline-operator ${OPERATOR}
```
nikhil-thomas, line 8: i understand why this change is required. But at present this will upset our ci systems and other operator-sdk commands which expect the repository root name to be same as the operator name.

As a workaround, you can check out this repository as `openshift-pipelines-operaotr`
```
git clone git@github.com:openshift/tektoncd-pipeline-operator.git openshift-pipelines-operator
```

or you can create a symbolic link with the name `openshift-pipelines-operator` to tektoncd-pipeline-operator directory

could you remove this change from this patch.

pkg/controller/config/config_controller.go
```diff
@@ -103,6 +103,10 @@ func readAddons(mgr manager.Manager) (mf.Manifest, error) {
 	if err != nil {
 		return mf.Manifest{}, err
 	}
+	if runtime1.GOARCH == "ppc64le" || runtime1.GOARCH == "s390x" {
+                log.Info("skip installation of tektoncd/catalog tasks if the platform is not x86_64")
```
nikhil-thomas, line 107: a statement would be better, instead of a conditional.

how about
```
log.Info("skipping installation of tektoncd/catalog tasks as the platform is not x86_64")
```

nikhil-thomas (changes requested):

nikhil-thomas:
@vdemeester @savitaashture @pradeepitm12

build/Dockerfile
pradeepitm12: Agree with @nikhil-thomas  !!!

snehlatamohite:
@nikhil-thomas - I have updated PR with suggested review comments. Please review.

PR #490, opened 2020-10-20
Update images tag to digest

savitaashture:
**Description:**

Disconnected cluster fails with imagePullBackoff error if images use tag but works well with digest, So modified images to use digest instead of tag

**ex:**
```
skopeo inspect docker://registry.redhat.io/rhel8/buildah | jq .Digest
"sha256:180c4d9849b6ab0e5465d30d4f3a77765cf0d852ca1cb1efb59d6e8c9f90d467"
```

/cc @khrm @piyush-garg @nikhil-thomas


vdemeester (approved):
/meow

PR #492, opened 2020-10-21
Introduce namespace exclusions for downgrading "pipeline" SA

sbose78:
### Goal

* Allow the admin to turn off `anyuid` in specific namespaces or all namespaces using a regex or "exact" string match

### What

* Added new API attribute called `spec.namespaceExclusions` in the `Config` CRD where the admin would be able to specify which namespaces are to have the `pipeline` SAs downgraded from `anyuid` to `restricted` by removing the rolebinding responsible for the same.

```
apiVersion: operator.tekton.dev/v1alpha1
kind: Config
spec:
  namespaceExclusions:  
    - sbose2                         ## regex is supported too.
  targetNamespace: openshift-pipelines
 ```
**Maybe support namespaceInclusions too?**

### What I've tested so far

1. Validate that the `Config` CR called `cluster` is auto-created ( no change )
2. Validate that a ClusterRole for "anyuid" is created.
3. Validate that all namespaces have `pipeline` sa bound to the above role using a rolebinding
4. Edit `cluster` to add specific namespaces to the list of `namespaceExclusions`
5. Validate removal of rolebindings seen in (2)
6. Empty the list of `namespaceExclusions` and see the rolebindings re-appear.

### TODO

- [x] Validate reconciliation of SAs, rolebindings, roles
- [x] Tests

### CODE REVIEW

**Comment by vdemeester** [comment]
@sbose78 I would like to split this into two.

One for:
* Downgraded controller sa from `privileged` to `nonroot`. 
* Granted `anyuid` to every pipeline sa. ( added a clusterrole and a rolebinding in every namespace )

A follow-up for 
* Added new API attribute called `spec.namespaceExclusions` in the `Config` CRD where the admin would be able to specify which namespaces are to have the `pipeline` SAs downgraded from `anyuid` to `restricted` by removing the rolebinding responsible for the same.

One reason for it is, so far, we do not put anything in the `Config` CR because it is automatically created without the user having anything to say about it (at install). I would like to experiment with http://openshift.github.io/openshift-origin-design/designs/administrator/olm/operand-creation-at-operator-install/ before putting some fields in the configuration.

File: pkg/controller/config/config_controller.go [hunk 1]
```diff
@@ -254,6 +264,38 @@ func (r *ReconcileConfig) Reconcile(req reconcile.Request) (reconcile.Result, er
...
+			}
+		}
+
+		if !downgraded {
+			err = r.upgradeDefaultSA(namespace)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+	}
```
nikhil-thomas, line 297: shall we move this logic and sub function in the the `rbac controller` https://github.com/openshift/tektoncd-pipeline-operator/blob/74c0817d8c2ebaf98622e0994010c38696b4876d/pkg/controller/rbac/rbac_controller.go#L58

the rbac controller watches `namespaces` so it might be possible to achieve `downgradeSA/upgradeSA` without this `for loop`.

sbose78: I did consider that, but the event would be on the `Config` resource and not on the `namespace` 🤔

pkg/controller/config/config_controller.go
```diff
@@ -91,13 +96,16 @@ func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
...
+	kc, _ := k8s.NewForConfig(mgr.GetConfig())
+
 	return &ReconcileConfig{
 		client:    mgr.GetClient(),
 		scheme:    mgr.GetScheme(),
 		pipeline:  pipeline,
 		triggers:  triggers,
 		addons:    addons,
 		community: community,
+		kc:        kc,
```
nikhil-thomas, line 108: are the type clients necessary.
could we use the dynamic client as it is already in `ReconcileConfig`.

pkg/controller/config/config_controller.go
sbose78: Having said that the rbac controller too needs to check ..
1. when a new namespace shows up.
2. create SA with anyuid only if not on the exclusion list.

sbose78:
@vdemeester Sounds good, have them in a single branch because I wanted to test it end-to-end  💇‍♂️ 

>  I would like to experiment with http://openshift.github.io/openshift-origin-design/designs/administrator/olm/operand-creation-at-operator-install/ before putting some fields in the configuration.


Link is broken :( or maybe I don't have access to it.

vdemeester:
image

Not sure why you don't have access though :stuck_out_tongue:

sbose78:
Magically the link works now ;) 
I didn't get you though 🤔 . 

I was hoping the admin would open the  `cluster` resource and edit it to exclude specific namespaces if she wants to. What is your proposed approach to specifying this information? 

🚗

vdemeester:
> I was hoping the admin would open the `cluster` resource and edit it to exclude specific namespaces if she wants to. What is your proposed approach to specifying this information?

Yeah that's a possibility. I don't necessarly have any other alternative right now.

sbose78:
Removed code which has been incorporated in previous PRs.

pkg/controller/config/config_controller.go
sbose78: The dynamic client wasn't working 🤔 . Wondering that's why the dynamic client wasn't used in the first place in the other controller 🤔 I can give it another try if you tell me :)

sbose78:
Somehow 4.4-e2e fails 

/retest

vdemeester (approved):
/cc @nikhil-thomas

gabemontero:
> @sbose78 I would like to split this into two.
> 
> One for:
> 
>     * Downgraded controller sa from `privileged` to `nonroot`.

hey @vdemeester  @sbose78 ... following up on this area after the holidays.

I know @vdemeester 's comment above was a ways back (Oct 2020). but given some of the user interaction just before the holidays (i.e. https://github.com/openshift/tektoncd-pipeline-operator/issues/545#issuecomment-749570146 for example) I'm curious if 

a) we need to resuscitate this PR and get it completed
b) if we should take things a step further on @vdemeester 's point and downgrade the controller/webhook SA's to `restricted`

I just tried it today with the master branch of tektoncd/pipelines and things work (at least the basics) with `restricted` ... see https://github.com/tektoncd/pipeline/issues/3625#issuecomment-754093515 for details.

thanks

> 
>     * Granted `anyuid` to every pipeline sa. ( added a clusterrole and a rolebinding in every namespace )
> 
> 
> A follow-up for
> 
>     * Added new API attribute called `spec.namespaceExclusions` in the `Config` CRD where the admin would be able to specify which namespaces are to have the `pipeline` SAs downgraded from `anyuid` to `restricted` by removing the rolebinding responsible for the same.
> 
> 
> One reason for it is, so far, we do not put anything in the `Config` CR because it is automatically created without the user having anything to say about it (at install). I would like to experiment with http://openshift.github.io/openshift-origin-design/designs/administrator/olm/operand-creation-at-operator-install/ before putting some fields in the configuration.

sbose78:
> if we should take things a step further on @vdemeester 's point and downgrade the controller/webhook SA's to restricted

That sounds good to me. I was rather playing safe, but since you've verified, I'm happy to make it restricted. The pipeline sa anyway would have the higher privileges for `buildah push`

gabemontero:
> > if we should take things a step further on @vdemeester 's point and downgrade the controller/webhook SA's to restricted
> 
> That sounds good to me. I was rather playing safe, but since you've verified, I'm happy to make it restricted. 

Cool deal

> The pipeline sa anyway would have the higher privileges for `buildah push`

Yep exactly.  And if we discover some other operation besides buildah push typically done in user pipelines needs something higher that `restricted`, they should get the necessary perms from using the `pipeline` SA

vdemeester:
> a) we need to resuscitate this PR and get it completed

Yes we should but it should be "re-done" on [`tektoncd/operator`](https://github.com/tektoncd/operator) now 🙃.

> b) if we should take things a step further on @vdemeester 's point and downgrade the controller/webhook SA's to `restricted`
> 
> I just tried it today with the master branch of tektoncd/pipelines and things work (at least the basics) with `restricted` ... see [tektoncd/pipeline#3625 (comment)](https://github.com/tektoncd/pipeline/issues/3625#issuecomment-754093515) for details.

Nice, I am all for that, let's wait for 1.3 to be released and let's do that 🙃

sbose78:
> Yes we should but it should be "re-done" on tektoncd/operator now 🙃.

Will do, Vincent!

saprette:
@sbose78  The SA pipeline is granted a role binding to the cluster role edit here
https://github.com/openshift/tektoncd-pipeline-operator/blob/2c35616a361ef2bb17a31e90bfc9995d9d6eddc6/pkg/controller/rbac/rbac_controller.go#L197

Is there any plan to make this configurable ? We would like to be able to restrict the resources that can be worked on in namespaces and granting the creation of pipeline/task/run indirectly opens pretty much everything without choice on the granularity.

namloc2001:
@gabemontero @vdemeester @sbose78 great continued conversation on this and related issues.

So from having the latest comments, does this mean that the change to actually be implemented is (at a high level): "downgrade the controller/webhook SA's to `restricted`".

And that because additional SCC access can be granted by SAs attached to the pipeline, the `namespaceExclusions` change isn't required?

gabemontero:
> @gabemontero @vdemeester @sbose78 great continued conversation on this and related issues.
> 
> So from having read the latest comments, does this mean that the change to actually be implemented is (at a high level): "downgrade the controller/webhook SA's to `restricted`".

Yes one of the upcoming changes (@vdemeester said he wanted to pursue it I believe after v1.3 of openshift pipelines goes out) is to assign the `restricted` SCC to the various tekton controller/webhook SAs

That does not mean  that is the only change that will be pursued.  See below.

> 
> And that because additional SCC access can be granted by SAs attached to the pipeline, the `namespaceExclusions` change isn't required?

I think there is still a desire to facilitate which SCCs an administrator would like assigned to the `pipeline` SA openshift pipelines provides out of the box.  Among other things, different configurations of image building tools, like buildah, can require different SCCs (anyuid vs. privileged).  Also, there is ongoing work in buildah that potentially can reduce the permissions needed, so we want flexibility there down the road.

Otherwise, if I'm not up to date on the latest nuances, I'm sure @vdemeester and @sbose78 will clarify.

PR #503, opened 2020-10-29
Run tekton controller as nonroot

sbose78:
Also renamed the rbac objects, wondering what's the impact beyond the fact that the old objects would lie around ?


vdemeester (approved):
/cc @nikhil-thomas

piyush-garg (approved):

nikhil-thomas:
> Also renamed the rbac objects, wondering what's the impact beyond the fact that the old objects would lie around ?

there should not be any other impact.

> `s/privileged-scc-role/pipelines-scc-role`

good lesson on naming :+1: :smile: 

/approve
/lgtm

PR #544, opened 2020-12-09
Remove enable repo in Dockerfile

piyush-garg:
This will remove --enablerepo=centosplus in yum install step
as the build is failing with unknown repo centosplus error

image builds is working without this also

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -10,7 +10,7 @@ ENV GIT_COMMITTER_EMAIL pipelines-dev@redhat.com
 
 RUN yum install epel-release -y
 
-RUN yum --enablerepo=centosplus install -y  \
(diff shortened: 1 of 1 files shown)


openshift-ci/Dockerfile.tools
```diff
@@ -10,7 +10,7 @@ ENV GIT_COMMITTER_EMAIL pipelines-dev@redhat.com
 
 RUN yum install epel-release -y
 
-RUN yum --enablerepo=centosplus install -y  \
```
savitaashture, line 13: do we need to enable any other repo instead `centosplus` ?

piyush-garg: this works without any repo, so I just removed it

nikhil-thomas:
/approve
/lgtm

it should automerge when the ci passes

PR #546, opened 2020-12-14
Skip installation of upstream tasks for ppc64le and s390x

snehlatamohite:
Hi All,

Done changes as per requested in https://github.com/openshift/tektoncd-pipeline-operator/issues/478#issuecomment-741531695

Please review. 

Snehlata Mohite.

PR #549, opened 2020-12-15
Skip upstream tasks when arch is P and Z, fix to branch v1.2.x

snehlatamohite:
Hi All,

As requested in https://github.com/openshift/tektoncd-pipeline-operator/pull/546 , updated code to skip installation of upstream tasks for P and Z.

Please review. 

Thanks,
Snehlata Mohite.

### DIFF
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
-
+        goruntime "runtime"
 	"github.com/go-logr/logr"
 	mfc "github.com/manifestival/controller-runtime-client"
 	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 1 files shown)


pkg/controller/config/config_controller.go
```diff
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
-
+        runtime1 "runtime"
```
vdemeester, line 10: 🤔

snehlatamohite: @vdemeester  - Hey I while building operator code observed below error hence I named runtime to runtime1. 

# github.com/tektoncd/operator/pkg/controller/config
pkg/controller/config/config_controller.go:107:5: undefined: "k8s.io/apimachinery/pkg/runtime".GOARCH
make: *** [out/operator] Error 2

snehlatamohite:
@vdemeester  - Ad discussed with Nikhil I have updated PR. There is a package conflict between
"k8s.io/apimachinery/pkg/runtime" and the standard library runtime and runtime1 is not a great naming choice so changed name to goruntime.

vdemeester (approved):
/lgtm

PR #580, opened 2021-02-19
Enable usage of tektoncd catalog tasks for s390x and ppc6le

barthy1:
The catalog tasks addition was disabled for `s390x` and `ppc6le` architectures because tasks themselves failed for non Intel architectures.
Now all tasks from the list can be executed on `s390x` and `ppc64le` (in some case container images are multi-arched, for other cases there is special parameter to specify custom image specific to architecture).
For several tasks version has to be bumped from `0.1` to `0.2`, because custom image parameter is supported only in version `0.2`.

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

### DIFF
--- a/pkg/controller/config/config_controller.go
+++ b/pkg/controller/config/config_controller.go
@@ -22,7 +22,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"os"
 	"path/filepath"
-	goruntime "runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


nikhil-thomas:
@barthy1 hi, this repository is archived. could you make this patch on tektoncd/operator.
TP-1.3 is based on this new repository.

ref: https://github.com/openshift/tektoncd-pipeline-operator#tektoncd-operator

barthy1:
@nikhil-thomas thanks for the info! I will create new PR to the proper repository

PR #633, opened 2021-06-15
Add a check for an existing sync PR in the midstream repo

alanfx:
This will allow the Jenkins jobs to use the actual return value of the
script, instead of grepping through the output.


nikhil-thomas:
@alanfx this repository is archived.
i have [removed the jobs from openshift-ci ](https://github.com/openshift/release/pull/19319)as well.
could you remove the jenkins triggering job.

so let us close this pr.

